PR #13, opened 2022-03-17
wip alt 2

oskarth:


rln/src/merkle.rs
```diff
@@ -1,26 +1,36 @@
...
     pub current_index: usize,
     merkle_tree: MerkleTree<E>,
 }
 
+// XXX: I don't understand this second where clause
 impl<E> IncrementalMerkleTree<E>
 where
-    E: Engine,
+    E: PairingEngine,
+    <<E as PairingEngine>::Fr as FromStr>::Err: Debug,
```
D4nte, line 27: `Fr` and `Err` are _associated types_ to `PairingEngine` and `FromStr` traits: https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html

the `as` terminology enables you to refer to said associated types.
At the end of the day, what this clause is trying to do is that `Err` is `Debug` which can then enable to debug print any `Err` in this `impl` block.

The `Err` we are talking about is the `Err` type associated to the `FromStr` trait so it has to do `< ..as FromStr>::Err` to specify that we are binding `FromStr::Err`. And then again to specify that we are binding `PairingEngine::Fr`.

rln/src/merkle.rs
```diff
@@ -1,26 +1,36 @@
...
 {
+    // XXX Some kind of trait bound thing?
     pub fn empty(hasher: Hasher<E>, depth: usize) -> Self {
         let mut zero: Vec<E::Fr> = Vec::with_capacity(depth + 1);
+        // XXX I don't understand why I can't call from_str here
+        // XXX minimal example?
         zero.push(E::Fr::from_str("0").unwrap());
```
D4nte, line 34: you need to specify that you want to use `Fr` as a `FromStr`. Something like `E::<Fr as FromStr>::from_str` might work.

oskarth:
Outdated

Issue #16, opened 2022-03-18
RLN: Make test_end_to_end verification succeed


oskarth:
# Problem

Currently https://github.com/vacp2p/zerokit/blob/master/rln/src/lib.rs#L51test is ignored because verification fails:

```
        // thread 'test::test_end_to_end' panicked at 'called `Result::unwrap()`
        // on an `Err` value: SynthesisError(MalformedVerifyingKey)',
        // rln/src/lib.rs:62:84
        //
        // Not sure why this is MalformedVerifyingKey, though the proof is
        // likely incorrect with wrong fields in protocol.rs
        //
        // Indeed:
        // if (public_inputs.len() + 1) != pvk.vk.gamma_abc_g1.len() {
 ```
 
 This requires changing `protocol.rs` (https://github.com/vacp2p/zerokit/blob/master/rln/src/protocol.rs#L85) so public inputs etc are correct.
 
 This should make tests pass, but it is also possible that the verification key from `zkey` file isn't correctly generated for some reason.
 
 ## Note
 
 This is using a different code path from JSON generate proof.
 
 This is a slightly more specific version of https://github.com/vacp2p/zerokit/issues/10
 
 ## Acceptance criteria
 
 End to end tests pass and RLN proof can be verified.

oskarth:
Some notes here: https://github.com/vacp2p/zerokit/commit/ddb3efc4a58a2efd9aac0765d905774fdfd2e2fd

I noticed the previous JSON-prepared proof (really witness) verified correctly, without calculated `y_share` etc. Not quite sure why that is, might be a bug.

PR #18, opened 2022-06-01
RLN: fix proof generation/verification and E2E tests

s1fr0:
This WIP PR addresses https://github.com/vacp2p/zerokit/issues/16 solving previous issues with proving and verifying keys import, proof generation/verification, RLN witness fields, etc. It further removes some unnecessary crates.

### DIFF
--- a/rln/Cargo.toml
+++ b/rln/Cargo.toml
@@ -65,7 +65,9 @@ blake2 = "0.8.1"
 sapling-crypto = { package = "sapling-crypto_ce", version = "0.1.3", default-features = false }
 bellman = { package = "bellman_ce", version = "0.3.4", default-features = false }
 
-semaphore = { git = "https://github.com/oskarth/semaphore-rs" }
+#semaphore = { git = "https://github.com/oskarth/semaphore-rs" }
+semaphore = { git = "https://github.com/worldcoin/semaphore-rs", rev = "d462a43"}
+
 
 tempfile = "3.3.0"
 
--- a/semaphore/Cargo.toml
+++ b/semaphore/Cargo.toml
@@ -21,7 +21,7 @@ num-bigint = { version = "0.4", default-features = false, features = ["rand"] }
 once_cell = "1.8"
 primitive-types = "0.11.1"
 rand = "0.8.4"
-semaphore = { git = "https://github.com/worldcoin/semaphore-rs" }
+semaphore = { git = "https://github.com/worldcoin/semaphore-rs", rev = "d462a43"}
 serde = "1.0"
 thiserror = "1.0.0"
 wasmer = { version = "2.0" }


rln/src/circuit.rs
```diff
@@ -33,24 +32,18 @@ pub fn VK() -> VerifyingKey<Bn254> {
...
 }
 
 pub fn CIRCOM() -> CircomBuilder<Bn254> {
-
     // Load the WASM and R1CS for witness and proof generation
-    let cfg = CircomConfig::<Bn254>::new(
-        WASM_PATH,
-        R1CS_PATH,
-    ).unwrap();  // should be )?; but need to address "the trait `From<ErrReport>` is not implemented for `protocol::ProofError`"
+    let cfg = CircomConfig::<Bn254>::new(WASM_PATH, R1CS_PATH).unwrap(); // should be )?; but need to address "the trait `From<ErrReport>` is not implemented for `protocol::ProofError`"
```
oskarth, line 46: I think if you run cargo `clippy` or `fmt` it will revert this

oskarth (commented):
What specifically didn't work before but works now? Not obvious from looking at code.

Also please run `cargo clippy` and `cargo fmt` as a habit, that way diff is more clean and format is uniform.

rln/src/lib.rs
```diff
@@ -87,54 +86,64 @@ mod test {
...
+        // We generate all relevant keys
+        let provingKey = &ZKEY();
+        let verificationKey = &VK();
+        let mut builder = CIRCOM();
+
+        // Let's generate a zkSNARK proof
+        let (proof, inputs) = generate_proof(builder, provingKey, rlnWitness).unwrap();
+
+        // Let's verify the proof
+        let success = verify_proof(verificationKey, proof, inputs).unwrap();
```
oskarth, line 128: Why did the API change here?

rln/src/main.rs
```diff
@@ -6,33 +6,17 @@ use ark_std::rand::thread_rng;
...
-//use tracing::{event, span, Level};
 use tracing_subscriber::layer::SubscriberExt;
 
 // JSON
 use serde::Deserialize;
-//use serde_json;
-
-#[derive(Debug, Deserialize)]
-//#[serde(rename_all = "camelCase")]
-struct WitnessInput {
```
oskarth, line 26: Sure we don't need this anymore?

s1fr0: I moved the data structure in protocol.rs https://github.com/vacp2p/zerokit/blob/6d3571034df74e786102c06cdf4e1b78a4015094/rln/src/protocol.rs#L31 since there can be populated with some helper functions with no need to expose all fields.

rln/src/lib.rs
s1fr0: I thought is better to standardize the APIs within each submodule in zerokit, abstracting the required inputs. For generating proof: the circom builder, a proving key and a witness. For verifying proofs the verification key, the proof and the circuit inputs to be verified.

In this way, for each submodule we need only to take care about generating witnesses, while the function signatures (modulo the type of the witness of course) will remain all the same (and do not depend on the module used). 

We can of course discuss the best approach here, not hard to change back.

s1fr0:
@oskarth As we already discussed, this implementation comes from implementing from scratch RLN using semaphore primitives for Merkle tree and Poseidon hash and checking outputs at each step against zk-kit RLN. Then I integrated only the required code to make test_end_to_end pass and now I'll proceed with eliminating the unnecessary parts. 
Thus regarding the consistency of the outputs generated, I cannot guarantee if there are differences or not, since I didn't check the output of [current master](https://github.com/vacp2p/zerokit/tree/8171439a440bc48870401e4ead31ca7a48186364), since RLN was already failing.

In general the main changes/fixes were about:
- Import of ZKEYs was somehow faulty. It is interesting that with [previous import function](https://github.com/vacp2p/zerokit/blob/8171439a440bc48870401e4ead31ca7a48186364/rln/src/circuit.rs#L14) a ZKEY was correctly imported and a proof computed, but then the verification always failed, even with a good input which passes verify when the zkey is randomly populated.
- Even when a ZKEY was correctly imported with the above fix (that is that passes verify over a good input), the implemented proof function `create_proof_with_reduction_and_matrices` seemed to fail with the matrices generated from `read_zkey`. It is still unclear to me what is the reason, but I suspect that the circom builder does some backend work by itself and [`create_random_proof_with_reduction`](https://github.com/vacp2p/zerokit/blob/6d3571034df74e786102c06cdf4e1b78a4015094/rln/src/protocol.rs#L214) was enough. This made random proofs created with imported proving ZKEY correctly verified.
- The rest are utility functions to import verification keys from JSON (useful for nodes to verify proofs with no need to have a proving key), verify them against proving keys (if present), and prepare RLN witnesses from JSON or by passing each fields.
- Other minor changes;

Note that [this](https://github.com/vacp2p/zerokit/pull/18/commits/6d3571034df74e786102c06cdf4e1b78a4015094) version already has `cargo clippy; cargo fmt` applied. I'll pipe these with cargo build/run so that are applied automatically.

rln/src/lib.rs
oskarth: OK, agree that'd be better, as long as we make the public API and C API match roughly what already exists, so should be fine

oskarth:
> Note that [this](https://github.com/vacp2p/zerokit/pull/18/commits/6d3571034df74e786102c06cdf4e1b78a4015094) version already has cargo clippy; cargo fmt applied. I'll pipe these with cargo build/run so that are applied automatically.

Oops my bad, I accidentally ended up reviewing a previous chunk.

Let me know ready for review and we can merge it

s1fr0:
@oskarth ready for review. I need to think a bit more on the C APIs, so I will update them in a followup issue.

rln/src/circuit.rs
```diff
@@ -1,28 +1,134 @@
...
+}
+
+// Utilities to convert a json verification key in a groth16::VerificationKey
+fn fq_from_str(s: &str) -> Fq {
+    BigInteger256::try_from(BigUint::from_str(s).unwrap())
+        .unwrap()
+        .into()
+}
+
+fn json_to_g1(json: &Value, key: &str) -> G1Affine {
```
oskarth, line 54: Maybe a comment for this function? Not obvious what this is about

rln/src/lib.rs
```diff
@@ -25,116 +20,108 @@ mod test {
...
+        let mut tree = PoseidonTree::new(tree_height, default_leaf);
+        tree.set(leaf_index, id_commitment.into());
 
-        println!("Root: {:#}", root);
-        println!("Merkle proof: {:#?}", merkle_proof);
+        // We check correct computation of the root
+        let root = tree.root();
+        assert_eq!(
+            root,
+            Field::from_str("0x27401a4559ce263630907ce3b77c570649e28ede22d2a7f5296839627a16e870")
```
oskarth, line 48: Is this from somewhere specifically?

rln/src/protocol.rs
```diff
@@ -103,121 +171,90 @@ pub enum ProofError {
...
 /// necessarily mean the proof is incorrect.
 pub fn verify_proof(
-    root: Field,
-    nullifier_hash: Field,
-    signal_hash: Field,
-    external_nullifier_hash: Field,
-    proof: &Proof,
+    verifying_key: &VerifyingKey<Bn254>,
+    proof: Proof,
+    inputs: Vec<Fr>,
```
oskarth, line 252: Why is this <Fr> instead of Field?

I don't quite see the purpose of removing the named parameters here as that removes context, ark_verify_proof is already general, the purpose of this function is to be specific to RLN AFAICT

rln/src/protocol.rs
```diff
@@ -103,121 +171,90 @@ pub enum ProofError {
...
-//
-
-// TODO Change API here
-/// Generates a semaphore proof
+/// Generates a RLN proof
 ///
 /// # Errors
 ///
 /// Returns a [`ProofError`] if proving fails.
 pub fn generate_proof(
```
oskarth, line 179: I'm still not sure I follow rationale of changing API here - the general API is `create_random_proof_with_reduction` which this specific function is hiding.

The RLNWitnessInput composite data structure seems OK to have in general, but I don't know why types are changed from Field to String etc.

Adding proving key here as requirement might be OK, depends a bit on how we imagine using this.

rln/src/protocol.rs
```diff
@@ -103,121 +171,90 @@ pub enum ProofError {
...
-        (
-            name.to_string(),
-            values.iter().copied().map(Into::into).collect::<Vec<_>>(),
-        )
-    });
-
+    mut builder: CircomBuilder<Bn254>,
+    proving_key: &ProvingKey<Bn254>,
+    rln_witness: RLNWitnessInput,
+) -> Result<(Proof, Vec<Fr>), ProofError> {
```
oskarth, line 183: What is in Vec<Fr> output? inputs? aka public outputs? Can be clarified

rln/src/public.rs
```diff
@@ -262,26 +122,3 @@ impl Default for RLN {
         Self::new()
     }
 }
-
-// NOTE: Expensive test, ignoring by default
-#[ignore]
-#[test]
-fn rln_proof() {
```
oskarth, line 269: This test was removed, I can't find the equivalent test that replicates this

oskarth (commented):
Good in general! See comments above

rln/src/lib.rs
s1fr0: No, was randomly generated by zk-kit implementation.

rln/src/protocol.rs
s1fr0: WitnessInput before had string fields https://github.com/vacp2p/zerokit/blob/master/rln/src/main.rs#L26, I just kept this. But I agree that it is better to have them another time and manage serialization on the witness data structure.

Since builder takes as inputs BigInts when pushing fields, probably this would be the best type for witness fields, so that we can avoid some conversions on the proof generation.

I agree that the general API is `create_random_proof_with_reduction`. We can either:
- preprocess builder outside and have a RLN specific proc to fill the builder and use the general API `create_random_proof_with_reduction` which takes as input builder and proving key;
- have a RLN specific `generate_proof` wrapping `create_random_proof_with_reduction` that feeds the witness into the builder and computes a proof.

In both case we will have at least 1 RLN specific procedure, unclear to me which approach is better.

rln/src/public.rs
s1fr0: You're right, is in the main now, will move.

oskarth (approved):
I'm going to approve it because it simplifies integration. There are some details we might have to iterate on but this is a good start. Nice work on your first Rust PR!

oskarth:
Once test/lint succeeds and you are happy with it, feel free to merge. Then we can open smaller PRs for specific improvements.

s1fr0:
Thanks @oskarth ! There are some other improvements implemented in this PR, which I would like to keep track of here:

- I standardized all primitives to work with semaphore's Field type, a wrapper over U256;
- Added some utility conversion from/to bytes, (bn254) Fr, and Field.
- Arithmetic over Field element is done in arkworks algebra by mapping elements to Fr and doing operation there efficiently.
- Added procedure to generate circuit's public inputs/outputs (e.g., x and y shares, nullifier, tree root, etc.) from witness. These values are stored in a new serializable struct called `RLNProofValues` which is passed to `verify_proof`.

These changes go towards https://github.com/vacp2p/zerokit/issues/19.

rln/src/protocol.rs
s1fr0: You're right. I created a new structure which is RLNProofValues and I now pass this to verify_proof instead of inputs: Vec<Fr>. This goes into the direction to have generate_proof and verify_proof have a general API, where to the first one you pass a module-specific witness (RLNWitness) while to the second proof-values to be verified (RLNProofValues).

rln/src/protocol.rs
s1fr0: Now this is removed, since users should not compute public circuit's input/output values evaluating witnesses in circuit. Utility functions will now do it directly from witnesses and generate_proof now only returns a zkSNARK proof.

Issue #19, opened 2022-06-03
RLN: Update API interface


s1fr0:
# Problem

PR https://github.com/vacp2p/zerokit/pull/18 solves https://github.com/vacp2p/zerokit/issues/16 (and https://github.com/vacp2p/zerokit/issues/10), but public APIs need to be updated to reflect implemented fixes/changes.

## Acceptance Criteria
- [x] Update [currently implemented](https://github.com/vacp2p/zerokit/tree/cf6168bf4e197246d99af45865df727dd016bb58) public APIs to work with fixes/changes introduced by https://github.com/vacp2p/zerokit/pull/18.

PR #20, opened 2022-06-14
RLN: Update Rust and C API

s1fr0:
This PR addresses https://github.com/vacp2p/zerokit/issues/19.

## Progress
- [x] Implement byte serialization/deserialization of custom data structure (RLN Witness and Proof values);
- [x] Extend/update RLN Rust public API to support all required features;
- [x] Extend/update RLN C public API to support all required features.


oskarth:
Just pinging @D4nte and @kaiserd in case you guys have any relevant experience with serialization like this in Rust and can spot immediate issues. If not that's fine :)

s1fr0:
Thanks @oskarth ! Besides the fact that byte serialization utility functions can be nicely implemented extending the Field/Fr type definition (this was mainly a PoC), I chose to serialize to `Vec<u8>` because I find it convenient for testing public rust API with Cursors  (rather than using [u8]), which implement Read/Write and I can then test outputs without writing to files or slicing. There might be better ways though.

rln/src/protocol.rs
```diff
@@ -25,13 +25,18 @@ use serde::{Deserialize, Serialize};
...
+    str_to_field, vec_field_to_bytes_be, vec_field_to_bytes_le, vec_to_field, vec_to_fr,
+    vec_u8_to_bytes_be, vec_u8_to_bytes_le,
+};
 
 ///////////////////////////////////////////////////////
 // RLN Witness data structure and utility functions
 ///////////////////////////////////////////////////////
 
-#[derive(Debug, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
kaiserd, line 39: Seems like this does not have to be clone. Is clone necessary for something planned?
Also, the serde traits Serialize and Deserialize are not used.

rln/src/protocol.rs
```diff
@@ -53,6 +58,98 @@ pub struct RLNProofValues {
     rln_identifier: Field,
 }
 
+pub fn serialize_witness(rln_witness: RLNWitnessInput) -> Vec<u8> {
```
kaiserd: Here, we could integrate that with Serde instead of writing a "stand-alone" function.

rln/src/protocol.rs
```diff
@@ -53,6 +58,98 @@ pub struct RLNProofValues {
...
+    serialized.append(&mut vec_field_to_bytes_le(rln_witness.path_elements));
+    serialized.append(&mut vec_u8_to_bytes_le(rln_witness.identity_path_index));
+    serialized.append(&mut field_to_bytes_le(rln_witness.x));
+    serialized.append(&mut field_to_bytes_le(rln_witness.epoch));
+    serialized.append(&mut field_to_bytes_le(rln_witness.rln_identifier));
+
+    serialized
+}
+
+pub fn deserialize_witness(serialized: Vec<u8>) -> RLNWitnessInput {
```
kaiserd: same

kaiserd (commented):
I left a few comments.
Generally, Serde is not used to its full extend.
Members of the `RLNWitnessInput` and `RLNProofValues` are deserialized using the Serde json deserialize, but the Structs themselfs are build in a "stand alone" function instead of implementing the respective Serde traits.

This is not a problem for functionally, but something that should be addressed for in the future for better maintainability.
(I could help here. Let's discuss in the Discord channel.)

For now, I did a shallow review, so take it with a grain of salt :sweat_smile:  I could do a more thorough review in a few days :).

D4nte:
> EDIT: since usize size might change based on architecture, when serializing vectors it is probably better to serialize its length not as a usize but rather as u32.

usize should be used when you are typing a memory address:

> The size of this primitive is how many bytes it takes to reference any location in memory. For example, on a 32 bit target, this is 4 bytes and on a 64 bit target, this is 8 bytes.

s1fr0:
> > EDIT: since usize size might change based on architecture, when serializing vectors it is probably better to serialize its length not as a usize but rather as u32.
> 
> usize should be used when you are typing a memory address:
> 
> > The size of this primitive is how many bytes it takes to reference any location in memory. For example, on a 32 bit target, this is 4 bytes and on a 64 bit target, this is 8 bytes.

Thanks @D4nte ! Indeed I use it to extract data chunks from a byte sequence/array. The problem I was concerned is if a vector (of some custom structure) is serialized within a 32 bits architecture and then deserialized over a 64 bits architecture: by keeping `usize` there might be some problems in the conversion since it has different sizes on these two machines. Also by keeping only `u32` there might happen undetected overflows. I then decided to store vector's size in a `u64` and use a try_form to detect any possible overflow when converting to `usize`, e.g. https://github.com/vacp2p/zerokit/blob/3d049df1ffc261e35d142703d08de4d5167326cd/rln/src/utils.rs#L211. There might be better ways though.

D4nte:
Why are you converting to `usize` in the first place? If I understand correctly, the bytes you are manipulating is arbitrary data. They are not pointers to memory location, hence you should **not** be using it at all.

You should only be using `Vec<usize>` if you decide to manipulate an array of memory addresses in a vector.

rln/src/utils.rs
```diff
@@ -1,34 +1,63 @@
...
-pub fn to_fr(el: Field) -> Fr {
-    Fr::try_from(el).unwrap()
+pub fn modulus_bit_size() -> usize {
+    <Fr as PrimeField>::Params::MODULUS
+        .num_bits()
+        .try_into()
+        .unwrap()
+}
+
+pub fn fr_byte_size() -> usize {
```
D4nte, line 16: Style comment. The usage of `fr`, `el` and other abbreviations feels very Fortran-esk and might not make the code as readable for the next developer or future you.

rln/src/utils.rs
```diff
@@ -1,34 +1,63 @@
...
+}
+
+pub fn to_fr(el: &Field) -> Fr {
+    Fr::try_from(*el).unwrap()
 }
 
-pub fn to_field(el: Fr) -> Field {
-    el.try_into().unwrap()
+pub fn to_field(el: &Fr) -> Field {
+    (*el).try_into().unwrap()
```
D4nte, line 26: Usage of `unwrap` is not recommended in production code, use `Result`. This utility function seems unnecessary.

s1fr0:
> Why are you converting to `usize` in the first place? If I understand correctly, the bytes you are manipulating is arbitrary data. They are not pointers to memory location, hence you should **not** be using it at all.
> 
> You should only be using `Vec<usize>` if you decide to manipulate an array of memory addresses in a vector.

My use case is to serialize `Vec<Foo>` from a serialization of `Foo`: the only missing information is the vector lenght, i.e. how many `Foo` objects I need to serialize/deserialize. Since at the end I will have a long byte vector, to deserialize i need to slice it based on the number of elements I expect to read and which i store at the beginning of the vector serialization (now) as a `u64`. To slice/iterate it seems I need the number of elements to be `usize`. I indeed might do a chunk read while loop exhausting the read u64 vector size integer and completely avoid the usize conversion.

D4nte:
> I indeed might do a chunk read while loop exhausting the read u64 vector size integer and completely avoid the usize conversion.

Makes sense, I missed that. I don't think chunk read would make sense because if you serialized a vector on a 64bit machine and the vector is longer than `uint32::max` then when the 32bit machine receives the data, it will not be able to put it in a `Vec` (ie, in memory) in the first place. So it will fail when the `Vec` is initialized at data reading and not when deserializing.

So in this specific instance,  I think it's ok to use `unwrap` (or event better, `expect`).

.gitmodules
```diff
@@ -1,5 +1,6 @@
 [submodule "rln/vendor/rln"]
 	path = rln/vendor/rln
+	ignore = dirty
```
oskarth, line 3: Did you make any relevant changes here?

rln/src/ffi.rs
```diff
@@ -51,13 +51,16 @@ pub extern "C" fn get_root(ctx: *const RLN, output_buffer: *mut Buffer) -> bool
 #[no_mangle]
 pub extern "C" fn new_circuit(ctx: *mut *mut RLN) -> bool {
     println!("rln ffi: new");
-    let mul = RLN::new();
+
+    let default_tree_height = 21;
```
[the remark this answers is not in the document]
oskarth: I think ti is fine if this is hardcoded for now, this corresponds to spec I believe. A comment could be nice tho

rln/src/ffi.rs
```diff
@@ -51,13 +51,16 @@ pub extern "C" fn get_root(ctx: *const RLN, output_buffer: *mut Buffer) -> bool
...
+    let mul = RLN::new(default_tree_height);
 
     unsafe { *ctx = Box::into_raw(Box::new(mul)) };
```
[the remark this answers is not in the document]
oskarth: Agree, at this point we are just trying to get something working, production-hardening can come later

rln/src/lib.rs
```diff
@@ -193,18 +192,33 @@ mod test {
...
-        let success = verify_proof(verification_key, proof, &proof_values).unwrap();
+        let success = verify_proof(&verification_key, &proof, &proof_values).unwrap();
 
         assert!(success);
     }
+
+    #[test]
+    fn test_serialization() {
+        // We test witness serialization
+        let rln_witness = rln_witness_from_json(WITNESS_JSON);
```
oskarth, line 213: this is fine, but I don't think we'll use the json? That was there as an example, it isn't part of the C API usage through nwaku (e.g.)

rln/src/protocol.rs
```diff
@@ -53,6 +61,97 @@ pub struct RLNProofValues {
     rln_identifier: Field,
 }
 
+pub fn serialize_witness(rln_witness: &RLNWitnessInput) -> Vec<u8> {
```
oskarth, line 64: Is this equivalent to https://github.com/kilic/rln? In terms of field order etc and usage through nwaku

oskarth (commented):
Nice, seems to be coming along :)

What are the next steps here? Do you see any issues with extending/updating the RLN C public API?

Could be useful to open a companion PR to nim-waku that exercises a subset of these API calls

s1fr0:
Thanks @oskarth ! I just pushed the C API so that you can have a look on the C API progress. Right now I'm finalizing tests for these C API and hopefully will push them today already.

rln/src/lib.rs
s1fr0: No, indeed we don't use it right now. But was useful in the debugging phase, so we might keep it in the codebase. Also in the future JSON may help exchanging data from different architecture/endpoints.

rln/src/protocol.rs
s1fr0: This is custom data structure, just the concatenation (LE by default) of the field values making a witness: 
```
identity_secret: Field,
path_elements: Vec<Field>,
identity_path_index: Vec<u8>,
x: Field,
epoch: Field,
rln_identifier: Field
```

Do you mean https://github.com/status-im/nwaku/blob/6c8ab0ab0f76fe765d95ef2f7bfb88d691a18bf7/waku/v2/protocol/waku_rln_relay/rln.nim#L48 ?

rln/src/main.rs
```diff
@@ -67,31 +53,28 @@ fn groth16_proof_example() -> Result<()> {
...
 
-    assert!(verified.unwrap());
+    let ser = serialize_proof_values(proof_values);
+    let b = deserialize_proof_values(ser);
+    println!("proof values deser: {:#?}", b);
 
     Ok(())
 }
 
 fn main() {
```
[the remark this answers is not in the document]
s1fr0: Main is now removed (was mainly used for testing).

rln/src/ffi.rs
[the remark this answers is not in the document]
s1fr0: Tree height is now a parameter.

rln/src/protocol.rs
s1fr0: Opened tracking issue https://github.com/vacp2p/zerokit/issues/21

rln/src/utils.rs
s1fr0: Thanks! FYI here fr indicates Fr, i.e. the finite field for scalars, so actually is not an abbreviation!

.gitmodules
s1fr0: Not really, I was annoyed by dirty submodule changes, and I saw that for semaphore there was this extra flag so I added it here and dirty submodules changes are now ignored.

rln/src/protocol.rs
s1fr0: Again comes from some unsuccessful tests... removed in https://github.com/vacp2p/zerokit/commit/6f0dda79854455945eb53e2f79ab12568701be88

rln/src/ffi.rs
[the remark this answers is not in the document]
oskarth: Let's break that up into an issue and tackle it separately, can be done after initial integration with nwaku is done

rln/src/lib.rs
oskarth: Ok agree, let's keep it in for testing and possibly future APIs

oskarth (approved):
Cool, nice work! Let's merge this and try. Will probably find bugs etc but easier to fix in follow up PRs once we start using it with nwaku

Issue #23, opened 2022-06-30
RLN: ark-circom CircomBuilder loads circuit resources slowly


s1fr0:
# Problem 

In zerokit RLN, [CircomBuilder](https://github.com/vacp2p/zerokit/blob/9cc333811b0a44638fe5ea5ac26a11f91a15aa68/rln/src/circuit.rs#L66-L75) from [ark-circom](https://github.com/gakonst/ark-circom) library is used to load the corresponding `.wasm` and `.r1cs` circuit files. 

This process may take up to 2-3 minutes at every RLN instance creation (witness generation and zk proof verification times are not affected)

# Acceptance criteria
- [x] Find out if this is expected behavior;
- [x] Minimize as much as possible loading time for circuit resources within ark-circom;
- [ ] If loading time are still not practical, implement alternative circuit loading mechanisms.

oskarth:
Are you and @richard-ramos both looking into this? Probably good to make sure there isn't overlap here

s1fr0:
It seems that compiling in `release` mode hugely improves performances. However, a better circuit logic, like the one implemented in https://github.com/vacp2p/zerokit/pull/26, further improves proof generation times.

s1fr0:
Issue resolved in https://github.com/vacp2p/zerokit/pull/26

PR #26, opened 2022-07-26
refactor(rln): improve circuit loading performances and tests fix

s1fr0:
This PR implements a different circuit loading mechanism -inspired by [semaphore-rs one's](https://github.com/worldcoin/semaphore-rs/blob/main/src/circuit.rs)- in order to address some performance issues described in https://github.com/vacp2p/zerokit/issues/23.

In particular, we replace the previous ark-circom CircomBuilder object used to load circuit and generate witnesses, with a witness generator obtained directly from the circuit .WASM file. As a result the .R1C1 is no more needed. These changes result in a more efficient zkSNARK proof generation.

This PR further fixes some tests for tree height equal 20 that were previously failing, due to missing/wrong hard-coded assert values.


rln/src/circuit.rs
```diff
@@ -8,34 +8,35 @@ use num_bigint::BigUint;
...
 use std::path::Path;
 use std::str::FromStr;
 
+use once_cell::sync::{Lazy, OnceCell};
+use std::sync::Mutex;
+use wasmer::{Module, Store};
+
 const ZKEY_FILENAME: &str = "rln_final.zkey";
 const VK_FILENAME: &str = "verifying_key.json";
-const R1CS_FILENAME: &str = "rln.r1cs";
```
oskarth, line 18: Is the r1cs file not useful for debugging? https://github.com/iden3/r1csfile/blob/master/doc/r1cs_bin_format.md

oskarth (approved):
Nice one! Only question is regarding debug information being lost when not allowing usage of r1cs file. Seems like a trade-offs that might look different in testing and production? Not sure

rln/src/circuit.rs
s1fr0: I expect most of the (circuit) debugging to be done in circom, not in zerokit. Here we just plug a verified/tested circuit compiled in circom and use it to generate proofs. If I get your concerns correctly, the debugging phase should happen before the compiled circuit is embedded in zerokit. Of course you might find some unexpected behaviors while using zerokit, but then it would most probably be a bad circuit implementation, hence you're back to circom. Does this address your concern?

oskarth: hmm, not sure

I think you are correct most debugging will happen at circom level, but I also see the need for debugging generally as well as for perf etc in Zerokit

but let's cross that bridge when we get there and ignore for now

Issue #28, opened 2022-07-29
RLN: initialize Merkle-tree passing height or number of leaves


s1fr0:
# Problem
In zerokit RLN module, a RLN object is initialized by passing the height of the membership Merkle-tree. The tree height corresponds to the total number of levels the tree is made of, including the leaves level.

In `nwaku` and in other implementations using RLN, it is common to initialize the Merkle-tree by passing the 2-logarithm of the total number of accumulatable leaves, that is an integer `N` in order to initialize a tree with `2^N` leaves.

However, a tree built over `2^N` leaves will have height `N+1`, hence the two parameters and initialization methods are not interchangeable.

# Acceptance criteria
- [ ] Decide which of the two above methods is more appropriate
- [ ] Align implementations (`zerokit` and/or `nwaku`)

@oskarth @staheri14

oskarth:
What does zk-kit and circomlibjs do?

PR #27, opened 2022-07-28
RLN: update README

s1fr0:
This PR updates the zerokit RLN module README in light of recent changes, by providing compilation instructions and commands to execute implemented tests.

PR #30, opened 2022-08-05
refactor(rln): Update Merkle tree initialization method

s1fr0:
This PR addresses https://github.com/vacp2p/zerokit/issues/28, by updating zerokit RLN Merkle tree implementation so that trees are initialized by passing the `tree_height = N` in order to be able to accumulate at most `2^N` hash leaves.

Previously, Merkle tree implementation followed [semaphore-rs](https://github.com/worldcoin/semaphore-rs/blob/d462a4372f1fd9c27610f2acfe4841fab1d396aa/src/merkle_tree.rs) one, and trees were initialized by passing the total number of node levels (including the pre-images of the hash leaves), i.e. `tree_height+1`.

As a consequence, previous circuits embedded in zerokit RLN, correspond to `height` equal to 15 and 19. Since `nwaku` requires a tree of height 20 for implementing RLN-RELAY, the corresponding circuit was added in the library and tests were updated to cover this case as well.

This PR affects https://github.com/vacp2p/zerokit/pull/27, and one between https://github.com/status-im/nwaku/pull/1030 and https://github.com/status-im/nwaku/pull/1060, mainly due to tests' hardcoded values that need to be updated.

### DIFF
--- a/rln/src/ffi.rs
+++ b/rln/src/ffi.rs
@@ -474,11 +474,10 @@ mod test {
         }
 
         if TEST_TREE_HEIGHT == 20 {
-            expected_path_elements.append(&mut vec![
-                Field::from_str(
-                    "0x1830ee67b5fb554ad5f63d4388800e1cfe78e310697d46e43c9ce36134f72cca",
-                )
-                .unwrap()]);
+            expected_path_elements.append(&mut vec![Field::from_str(
+                "0x1830ee67b5fb554ad5f63d4388800e1cfe78e310697d46e43c9ce36134f72cca",
+            )
+            .unwrap()]);
             expected_identity_path_index.append(&mut vec![0]);
         }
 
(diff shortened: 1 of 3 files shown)


rln/src/ffi.rs
```diff
@@ -450,8 +450,8 @@ mod test {
         let mut expected_identity_path_index: Vec<u8> =
             vec![1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
 
-        // We add the remaining elements for the case TEST_TREE_HEIGHT = 20
-        if TEST_TREE_HEIGHT == 20 {
+        // We add the remaining elements for the case TEST_TREE_HEIGHT = 19
+        if TEST_TREE_HEIGHT == 19 || TEST_TREE_HEIGHT == 20 {
```
oskarth, line 454: Why the || condition? shouldn't this only be for 19?

oskarth (approved):

rln/src/ffi.rs
oskarth: ah right, nvm

oskarth:
Lint failing otherwise lgtm

Issue #29, opened 2022-08-03
RLN: use space-efficient Merkle Tree implementation


s1fr0:
# Problem
[Currently](https://github.com/vacp2p/zerokit/commit/cbb10efc10adc4ae8d654c59e86740161669bd80), zerokit RLN supports naive Merkle Tree implementation: when a RLN object is created, a tree of height `N+1` is initialized (including the leaves level) with a total of `2^(N+1)-1` allocated nodes that remain allocated for the whole life of the RLN object, even in case the tree is empty or accumulates less than `2^N` leaves.

For space-efficiency reasons we would like to add support to a Merkle trees implementations that progressively allocates only those nodes which are necessary to prove membership of currently non-empty leaves.

# Acceptance criteria
- [x] Implement Incremental Merkle Tree with APIs compatible with current Merkle Tree implementation (aligned to https://github.com/vacp2p/zerokit/issues/28)
- [ ] ~~No regression in performances with respect to current implementation.~~
- [ ] ~~If regression in performance, evaluate space/time tradeoffs and followup with a better solution.~~

staheri14:
Just for reference and following our conversation in the PM channel, the Merkle tree implementation of the killic lib satisfies the requirement of this issue https://github.com/kilic/rln/blob/master/src/merkle.rs, feel free to use that

staheri14:
Not sure about this last item 
> If regression in performance, evaluate space/time tradeoffs and followup with a better solution.
But if I understand it correctly, this maps to this issue https://github.com/vacp2p/research/issues/109 which I am already working on it and the result of my work will be ultimately integrated into zero-kit.

oskarth:
To make sure I understand correctly, this issue is about porting the incremental MT algorithm already used in kilic.

1) Is this correct and there's nothing else in the scope of this?
- I believe the answer is yes 
2) How confident are we this code will be ultimately used once we've done more analysis in e.g. https://github.com/vacp2p/research/issues/109?
- Assuming same MT approach, my intuition would be fairly
- However long term I could imagine a different representation, in which case this solution would be an interim one
- It seems to me like an incremental MT representation plus possibly additional tweaks discovered would be "good enough" for cross-client test and for smaller user sets, i.e. for foreseeable future
3) Do we want this to be implemented before we switch to Zerokit by default? (after this testnet)
- Intuition says yes, but I'm not familiar enough about details of current perf problems for MT storage

oskarth:
[...]

PR #32, opened 2022-08-10
refactor(rln): use space-efficient Merkle tree implementation

s1fr0:
This PR addresses https://github.com/vacp2p/zerokit/issues/29 by adapting and integrating in a single space-efficient Merkle tree implementation:
- [Kilic's RLN](https://github.com/kilic/rln/blob/master/src/merkle.rs) Merkle Tree implementation;
- [semaphore-rs](https://github.com/worldcoin/semaphore-rs/blob/d462a4372f1fd9c27610f2acfe4841fab1d396aa/src/merkle_tree.rs) Merkle tree implementation.

In contrast to the previous zerokit Merkle tree implementation (mainly based on semaphore-rs' one), preliminary performance tests for the one proposed in this PR shows that, on average:
- an `update_next` call runs 4-5 microseconds **slower** (that is, ~4s overall delay for setting all leaves of a tree of depth 20);
- a new tree is initialized ~60x times faster;
- theoretically, setting N leaves should now require ~2^(log2(N)+1)*32 bytes instead of ~2^(N+1)*32;

The first two items can be verified running: `cargo test test_merkle_performance --release -- --nocapture`
However, the same test should be adapted to the previous zerokit implementation -now removed- to effectively compare results.

A thorough analysis on storage performance (third item) and possible optimizations will follow in https://github.com/vacp2p/research/issues/127.

### DIFF
--- a/rln/src/merkle_tree.rs
+++ b/rln/src/merkle_tree.rs
@@ -45,20 +45,20 @@ where
     H: Hasher,
 {
     /// The depth of the tree, i.e. the number of levels from leaf to root
-    pub depth: usize,
-    
+    depth: usize,
+
     /// The nodes cached from the empty part of the tree (where leaves are set to default).
     /// Since the rightmost part of the tree is usually changed much later than its creation,
     /// we can prove accumulation of elements in the leftmost part, with no need to initialize the full tree
     /// and by caching few intermediate nodes to the root computed from default leaves
     cached_nodes: Vec<H::Fr>,
-    
+
     /// The tree nodes
     nodes: HashMap<(usize, usize), H::Fr>,
-    
+
     // The next available (i.e., never used) tree index. Equivalently, the number of leaves added to the tree
     // (deletions leave next_index unchanged)
-    pub next_index: usize,
+    next_index: usize,
 }
 
 /// The Merkle proof
@@ -69,7 +69,6 @@ pub struct OptimalMerkleProof<H: Hasher>(pub Vec<(H::Fr, u8)>);
 /// Implementations
 
 impl<H: Hasher> OptimalMerkleTree<H> {
-
     pub fn default(depth: usize) -> Self {
         OptimalMerkleTree::<H>::new(depth, H::default_leaf())
     }
@@ -91,9 +90,19 @@ impl<H: Hasher> OptimalMerkleTree<H> {
         }
     }
 
+    // Returns the depth of the tree
+    pub fn depth(&self) -> usize {
+        self.depth
+    }
+
     // Returns the capacity of the tree, i.e. the maximum number of accumulatable leaves
     pub fn capacity(&self) -> usize {
         1 << self.depth
+    }
+
+    // Returns the total number of leaves set
+    pub fn leaves_set(&mut self) -> usize {
+        self.next_index
     }
 
     #[must_use]
@@ -116,7 +125,6 @@ impl<H: Hasher> OptimalMerkleTree<H> {
         Ok(())
     }
 
-    #[must_use]
     // Sets a leaf at the next available index
     pub fn update_next(&mut self, leaf: H::Fr) -> io::Result<()> {
         self.set(self.next_index, leaf)?;
@@ -127,17 +135,11 @@ impl<H: Hasher> OptimalMerkleTree<H> {
     pub fn delete(&mut self, index: usize) -> io::Result<()> {
         // We reset the leaf only if we previously set a leaf at that index
         if index < self.next_index {
-            self.set(index, H::default_leaf())?;            
+            self.set(index, H::default_leaf())?;
         }
         Ok(())
     }
 
-    // Returns the total number of leaves set
-    pub fn leaves_set(&mut self) -> usize {
-        self.next_index
-    }
-
-    #[must_use]
     // Computes a merkle proof the the leaf at the specified index
     pub fn proof(&self, index: usize) -> io::Result<OptimalMerkleProof<H>> {
         if index >= self.capacity() {
@@ -162,7 +164,6 @@ impl<H: Hasher> OptimalMerkleTree<H> {
         Ok(OptimalMerkleProof(witness))
     }
 
-    #[must_use]
     // Verifies a Merkle proof with respect to the input leaf and the tree root
     pub fn verify(&self, leaf: &H::Fr, witness: &OptimalMerkleProof<H>) -> io::Result<bool> {
         if witness.length() != self.depth {
@@ -212,7 +213,6 @@ impl<H: Hasher> OptimalMerkleTree<H> {
 }
 
 impl<H: Hasher> OptimalMerkleProof<H> {
-
     #[must_use]
     // Returns the length of a Merkle proof
     pub fn length(&self) -> usize {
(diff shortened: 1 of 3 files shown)


oskarth (commented):
Nice one! Some comments

I agree storage is probably more important. The 20s delay, when would this happen in e.g. current testnet?

Does it make sense to have both impl side by side or are we confident in replacing it?

Have we tried this with nwaku to make sure it still works?

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
-/// For a given node index, return the parent node index
-/// Returns None if there is no parent (root node)
-const fn parent(index: usize) -> Option<usize> {
-    if index == 0 {
-        None
-    } else {
-        Some(((index + 1) >> 1) - 1)
-    }
-}
+pub struct MerkleProof<H: Hasher>(pub Vec<(H::Fr, u8)>);
```
staheri14, line 49: An specific reason for not using `bool` instead of `u8`?

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
         }
-        if count != 0 {
-            self.update_nodes(index, index + (count - 1));
-            self.next_index = max(self.next_index, start + count);
+        let mut witness = Vec::<(H::Fr, u8)>::with_capacity(self.depth);
+        let mut i = index;
+        let mut depth = self.depth;
+        loop {
+            i ^= 1;
+            witness.push((self.get_node(depth, i), (1 - (i & 1)).try_into().unwrap()));
```
staheri14, line 114: `1` means right and `0` means left  but this computation `(1 - (i & 1))` does the reverse.  Below suggestion (which correspond to the killic lib) works if you use `bool` instead of `u8`. 
```suggestion
            witness.push((self.get_node(depth, i), ((i & 1 == 1)).try_into().unwrap()));
```

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
+            ));
+        }
+        let mut acc = *leaf;
+
+        for w in witness.0.iter() {
+            if w.1 == 0 {
+                acc = H::hash(&[acc, w.0]);
+            } else {
+                acc = H::hash(&[w.0, acc]);
             }
```
staheri14, line 139: Referring to this previous comment, this if-else blocks also needs to be swapped. This did not happen to error out because the logic was swapped symmetrically in both functions i.e., `verify` and `proof`. 
```suggestion
            if w.1 == 0 {
                acc = H::hash(&[w.0, acc]);
            } else {
               acc = H::hash(&[acc, w.0]);     
            }
```

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
 
     #[must_use]
-    pub fn get_path_elements(&self) -> Vec<H::Hash> {
-        self.0
-            .iter()
-            .map(|x| match x {
-                Branch::Left(value) | Branch::Right(value) => *value,
-            })
-            .collect()
+    pub fn get_path_elements(&self) -> Vec<H::Fr> {
```
staheri14, line 185: A function description would be helpful.

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
-where
-    H: Hasher,
-    H::Hash: Debug,
-{
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Self::Left(arg0) => f.debug_tuple("Left").field(arg0).finish(),
-            Self::Right(arg0) => f.debug_tuple("Right").field(arg0).finish(),
+    #[must_use]
+    pub fn root(&self, leaf: &H::Fr) -> H::Fr {
```
staheri14, line 205: Why this function is required? Isn't root always updated?

staheri14 (commented):
Looks good to me in general! left my comments inline.
I also think that it would be better to keep the previous MT implementation, there is no harm for both algorithm to co-exit. Once we decide which MT algorithm outperforms the other we may stick to one only. But even in that case, having both of them gives more options to the users of this repo.

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
+where
+    H: Hasher,
+{
+    /// The depth of the tree
+    pub depth: usize,
+    /// The nodes cached from the empty part of the tree (where leaves are set to default).
+    /// Since the rightmost part of the tree is usually changed much later than its creation,
+    /// we can prove accumulation of elements in the leftmost part, with no need to initialize the full tree
+    /// and by caching few intermediate nodes to the root computed from default leaves
+    cached_nodes: Vec<H::Fr>,
```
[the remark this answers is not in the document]
staheri14: Yes, I know the purpose of the `zero` vector, it essentially a way to represent a `zero` or `empty` Merkle tree with logarithmic storage complexity i.e., by only `d+1` elements. The name `zeros` also  mirrors how it is being called in the literature. It is fine to keep it `cached_nodes`, but sticking to the name `zeros` allows other developers to quickly understand it maps to the actual `zeros` algorithm and nothing different.

rln/src/merkle_tree.rs
```diff
@@ -1,251 +1,231 @@
...
+    pub fn set(&mut self, index: usize, leaf: H::Fr) -> io::Result<()> {
+        if index >= self.tree_size() {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "index exceeds set size",
+            ));
+        }
+        self.nodes.insert((self.depth, index), leaf);
+        self.recalculate_from(index);
+        self.next_index = max(self.next_index, index + 1);
```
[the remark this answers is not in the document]
staheri14: Oh, that gets complicated, such option makes sense in general use-cases but not for the RLN membership Merkle tree, right? do you see a specific use-case for such sparse leaf assignment? it might be easier to remove this flexibility i.e., the `set` function 
Also, another question would be why `N` up to `N+M-1` are left empty in this example?

rln/src/merkle_tree.rs
staheri14: Fine with adjusting it based on needs, but then this new indexing differs from the conventional indexing (where branches to the left are indexed by 0 and to the right by one). If we could stick to the conventional one, it makes the code more readable and compliant with what is commonly known in the literature. 
BTW, no rush to do it in this PR, another PR would be fine.

staheri14 (approved):
LGTM! (posted a few questions are inline)

rln/src/merkle_tree.rs
oskarth: > where branches to the left are indexed by 0 and to the right by one

Agree this is more natural. I don't quite understand the argument for not doing this?

rln/src/merkle_tree.rs
```diff
@@ -1,132 +1,382 @@
-// Implementation adapted from https://github.com/worldcoin/semaphore-rs/blob/d462a4372f1fd9c27610f2acfe4841fab1d396aa/src/merkle_tree.rs
-// In our customization, we expand MerkleTree to have a next_index counter, so that we can add a public API to add leaves to the next available counter with no need to specify the index
+// This crate provides different implementation of Merkle tree
+// Currently two interchangeable implementations are supported:
+//    - FullMerkleTree: each tree node is stored
+//    - OptimalMerkleTree: only nodes used to prove accumulation of set leaves are stored
```
oskarth, line 4: I'm confused by "optimal" name, where is this from? is this not just IncrementalMerkleTree? Or perhaps Sparse MT

oskarth (approved):

rln/src/merkle_tree.rs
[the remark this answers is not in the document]
oskarth: No strong opinions on above, a meta comment though: try to avoid back and forth in PRs, especially if isn't just asking about clarification but more of a discussion. If a choice is controversial it is better to make the merge the part that people agree on/are OK with, and isolate discussions to specific follow up issues.

Usually, it doesn't matter that much.

rln/src/merkle_tree.rs
oskarth: ah my bad, I confused incremental and optimal based on how I thought kilic used incremental mt algo

Issue #31, opened 2022-08-08
Milestone: zerokit release v0.1


s1fr0:
## Problem

**UPDATE: This is becoming a waterfall release which keeps getting delayed, which was never the intention. We want to take existing code and cut it into an initial 0.1 release.**

We want to publish a release v0.1 for Zerokit with a first ~~mature~~ version of the RLN module.

Before that, it is desirable to clean-up the RLN module library code and implement different optimizations which are already standard in other libraries.

These allow to: make the RLN module more efficient and easily integrable in projects employing the RLN primitive; bring a cleaner code and optimized cross-module primitives (e.g., Merkle trees) to new zerokit modules.

## Scope

Release zerokit v0.1 with a RLN module that can replace and improve Kilic's RLN module currently employed by `nwaku`.

**Must have**:
- [x] Sanity check main features included / bug fixes included
-- Err on side of not accepting more scope
- [x] Release branch with release note and tag
- [ ] Announce release on Discord and Twitter

**Initial list of issues, many nice to haves that can be dealt with separately**:
- [x] https://github.com/vacp2p/zerokit/issues/29
* [ ] Maximize usage of only arkworks dependencies (e.g., replace Semaphore's Field with arkworks Fr, etc.) and remove unnecessary ones (might require adapting/integrating used primitives e.g. Poseidon hash)
   - [x] https://github.com/vacp2p/zerokit/issues/35
   - [ ] https://github.com/vacp2p/zerokit/pull/44
* [ ] Eventually add/update APIs to address new needs resulting from completions of milestones in the RLN-RELAY track https://github.com/status-im/nwaku/issues/1062 or other tracks:
   - [ ] https://github.com/vacp2p/zerokit/pull/45 (contains https://github.com/vacp2p/zerokit/pull/36)
- [ ] Reduce size of compiled libraries (might require partial refactoring of dependencies) https://github.com/vacp2p/zerokit/issues/37

## Acceptance criteria

Issues defined in scope addressed.

## Risks and uncertainty

Removing some dependencies (e.g. semaphore-rs) might result not trivial, since there might some implementation used across the code base that should be ported to the replacing dependencies.

## Out of scope

This milestone's scope requires a RLN module which can replace (in terms of APIs and efficiency) Kilic's RLN module. However, integration of zerokit v0.1 in nwaku is out-of-scope and is tracked in https://github.com/status-im/nwaku/issues/1061.

## Notes and links

[...]

PR #34, opened 2022-08-22
WIP: New implementations and tests for Zerokit v0.1

s1fr0:
Targets https://github.com/vacp2p/zerokit/issues/31


richard-ramos (approved):
:100: 

Looking at the code, a random idea i had that wanted to ask you about is:
What do you think of extracting in a patch file or separate branch these changes?:
- in poseidon_tree.rs:
```rs
//pub type PoseidonTree = FullMerkleTree<PoseidonHash>;
//pub type MerkleProof = FullMerkleProof<PoseidonHash>;
```
- the functions defined in `rln/src/utils.rs`. 

That way the source code would not contain commented code, and someone wanting to use either the FullMerkleTree, or the conversion utils could either cherry pick the commit from the branch, or apply the git patch  (assuming this operation is documented somewhere)

richard-ramos:
Sounds good! I did not know about rust features!

oskarth (changes requested):
This is not a release PR, it is doing _a lot_ of things at once. These should be separate PRs. The release PR should be minimal and only about adding release notes etc.

See e.g. how this is done in nwaku: https://github.com/status-im/nwaku/pull/1075/files

I'm also not convinced that we should decrease the capability this library has as part of the first release.

multiplier/Cargo.toml
```diff
@@ -1,35 +0,0 @@
-[package]
```
oskarth, line 1: This is a useful example for onboarding and understanding roughly how things fit together. I don't think it makes sense to remove this package.

We could move it to an example directory.

poseidon-tornado/README.md
```diff
@@ -1,15 +0,0 @@
-# Poseidon-Tornado wrapper
```
oskarth, line 1: We want this soon, but I agree it is in a so-so state and better to base based on RLN from scratch at this point.

oskarth: Removed in https://github.com/vacp2p/zerokit/commit/1d8a5f59b60eea88cdd69d68df760ebb3ae73e54

cc @staheri14 if you want to see what this code is doing

rln/Cargo.toml
```diff
@@ -3,76 +3,39 @@ name = "rln"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
oskarth, line 6: Do this in a separate PR

rln/src/circuit.rs
```diff
@@ -1,22 +1,19 @@
 // This crate provides interfaces for the zero-knowledge circuit and keys
 
-use ark_bn254::{Bn254, Fq, Fq2, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
-use ark_circom::{read_zkey, CircomBuilder, CircomConfig, WitnessCalculator};
-use ark_ff::BigInteger256;
+use ark_bn254::{
+    Bn254, Fq as ArkFq, Fq2 as ArkFq2, Fr as ArkFr, G1Affine as ArkG1Affine,
```
oskarth, line 4: This a change in functionality or so, belongs in a separate PR targeting an issue. Not in a release PR...

rln/src/merkle_tree.rs
```diff
@@ -35,239 +32,6 @@ pub trait Hasher {
     fn hash(input: &[Self::Fr]) -> Self::Fr;
 }
 
-////////////////////////////////////////////////////////////
```
oskarth, line 38: This shouldn't be removed, it is a regression

rln/src/poseidon_hash.rs
```diff
@@ -0,0 +1,129 @@
+// This crate implements the Poseidon hash algorithm https://eprint.iacr.org/2019/458.pdf
+
+// The implementation is taken from https://github.com/arnaucube/poseidon-rs/blob/233027d6075a637c29ad84a8a44f5653b81f0410/src/lib.rs
+// and slightly adapted to work over arkworks field data type
+
```
oskarth, line 5: This seems like another feature ??

semaphore/Cargo.toml
```diff
@@ -1,49 +0,0 @@
-[package]
```
oskarth, line 1: I think this is still useful, we aren't unlikely to want to use this in the feature. There are no problems to keep it in the code base as WIP, and when it gets prioritized to update it.

oskarth:
Yeah I think better to close this PR and start from scratch with feature PRs and refactor PRs and then a minimal release note PR. This is just a monster right now with many separate intentions.

rln/Cargo.toml
```diff
@@ -3,76 +3,39 @@ name = "rln"
 version = "0.1.0"
```
oskarth, line 3: @richard-ramos @staheri14  of the changes in the `rln` crate, which of these are must haves from your POV? Both for normal integration (e.g. bugs found in nwaku and go-waku) as well as for the experimental wasm support?

Trying to figure out which things should be part of initial v0.1 release and which are OK to deal with later.

Issue #35, opened 2022-08-22
RLN: Use arkworks field arithmetic only


s1fr0:
# Problem
Currently, zerokit RLN uses both semaphore-rs `Field` and ark-ec `Fr` data types to carry field arithmetic, in order to use natively primitives from dependencies implemented for one type and the other. 

However, while the semaphore-rs `Field` type corresponds to a big-endian byte vector without Montgomery reduction, the arkworks `Fr` type does operations directly in the more efficient Montgomery form, hence modulo p reduction is implicit.

This issue partially overlaps with the second item of https://github.com/vacp2p/zerokit/issues/31: while, in the latter, we ask to maximize usage of the `Fr` type, here we aim to exclusively use `Fr` along the whole module. Such refactor might require much more time than what is expected for zerokit v0.1 release.

# Acceptance Criteria
- [x] Use exclusively the arkworks `Fr` field type to perform arithmetic in zerokit RLN

oskarth:
Can we please solve this in a normal PR and not bundle everything up into a mega PR for release.

PR #43, opened 2022-09-14
Use arkworks arithmetic only

s1fr0:
Implements https://github.com/vacp2p/zerokit/issues/35#issuecomment-1246183424


oskarth (approved):
Great, thanks!

PR #44, opened 2022-09-14
feat(RLN): integrate Poseidon Hash

s1fr0:
This PR integrates an adaptation of [poseidon-rs](https://github.com/arnaucube/poseidon-rs/blob/233027d6075a637c29ad84a8a44f5653b81f0410/src/lib.rs) Poseidon hash implementation into zerokit RLN that works directly over arkworks arithmetic, in use since last [refactor](https://github.com/vacp2p/zerokit/pull/43).


oskarth (approved):
Nice! I assume this has been tested for compatibility with existing circuits etc.

Issue #50, opened 2022-09-26
Generate Poseidon round parameters and constants


s1fr0:
# Problem
We would like to have a self-contained `poseidon` module that works natively with arkworks traits, so that round parameters and constants are generated directly from the field characteristic and not hard-coded as now are. This will further enable current and future zerokit modules that use it to access the same Poseidon implementation.

In order to achieve such goal, we can progressively refactor the poseidon implementation [currently](https://github.com/vacp2p/zerokit/tree/14f41d5340fefc0919c05273ae5c35b2b64b1a30) present in the `rln` module. 

The main required steps should be:
- [x] Implement `GrainLFSR` in order to generate constants from hard-coded round parameters (i.e., the tuples `(t, RF, RP)`) and field characteristic;
- [x] Pass arkworks `PrimeField` trait to Poseidon, so that the latter can be instantiated using different curves;
- [ ] Generate in-code round parameters, i.e. port [generate_params_poseidon.sage](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_params_poseidon.sage)
- [ ] Implement algorithm [1](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_parameters_grain.sage#L197)-[2](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_parameters_grain.sage#L238)-[3](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_parameters_grain.sage#L276) from reference implementation to check security of randomly generated MDS matrices. Before implementing these algorithms, we hard-code the number of randomly generated unsecure matrices that should be skipped during generation (value taken from reference implementation);
- [x] Move the `poseidon` implementation to a separate module, e.g. under `utils`. This step can be done at any moment.

# Acceptance Criteria
- [ ] Implement all the above steps.

PR #51, opened 2022-09-26
Add GrainLFRS and PrimeField trait to Poseidon

s1fr0:
Implements the first two steps of https://github.com/vacp2p/zerokit/issues/50.

### DIFF
--- a/rln/src/poseidon_constants.rs
+++ b/rln/src/poseidon_constants.rs
@@ -276,7 +276,7 @@ pub fn find_poseidon_ark_and_mds<F: PrimeField>(
 mod test {
     use super::*;
     use crate::circuit::Fr;
-    use crate::poseidon_hash::{Poseidon};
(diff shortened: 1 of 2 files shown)


oskarth (commented):
We need to audit this code at some point...

I can't give good review, but seems... reasonable?

> We would like to have a self-contained poseidon module that works natively with arkworks traits, so that round parameters and constants are generated directly from the field characteristic and not hard-coded as now are. This will further enable current and future zerokit modules that use it to access the same Poseidon implementation.

Can you elaborate a bit on a specific example task that becomes easier to do with this? Can't quite picture it right now.

oskarth:
The first point seems orthogonal to the adding of traits (unless I missed something?), but agree with the second point!

oskarth (approved):
LGTM assuming tested with nwaku e.g.

richard-ramos (approved):

Issue #49, opened 2022-09-26
Enable library consumer to pass their entropy source


fryorcraken:
## Problem

When generating credentials in the browser, one need to find alternative way to export/backup the credentials for safe keeping.

In the Ethereum/zk domain, it is current practice to use the hash of the signature of a specific message as entropy for RLN credentials.
This enables the user to retrieve their credentials by using the same wallet. Meaning their zk credentials are backed up with their HD wallet.

In zerokit, the API does not allow the library consumer to pass a seed or entropy source

## Proposed solution

Enabling passing a seed on the wasm API for credential generation. I'd expect the seed to be 32 bytes, please clarify if it's not the case.

PR #56, opened 2022-09-28
feat(rln): add seeded keygen

s1fr0:
Addresses https://github.com/vacp2p/zerokit/issues/49 by implementing in the `rln` module a `seeded_keygen` function to generate an `(id_secret, id commitment)` pair from a Chacha20 RNG seeded by the user.

Such API is implemented in both `public` and `ffi` crates. Tests are implemented as well.

The choice of using Chacha20 is to minimize the risk of obtaining different results among different platform and dependencies' versions.

ChaCha20 requires a seed of exactly 32 bytes: for this reason the (arbitrary long) input seed provided by the user is hashed using Keccak256 before being passed to ChaCha20 as seed. This enables to use seed phrases as well in order to generate identity pairs. Note however, that Keccak has very efficient hardware implementations and might not be the right choice to protect short seed phrases that would be then become vulnerable to brute-force attacks. 

Note: since a new API is added, to access it in `nwaku` we need to regenerate the RLN nim bindings. This will be done in a nwaku PR once this is merged.

### DIFF
--- a/rln/Cargo.toml
+++ b/rln/Cargo.toml
@@ -35,7 +35,7 @@ once_cell = "1.14.0"
 rand = "0.8"
 rand_chacha = "0.3.1"
 tiny-keccak = { version = "2.0.2", features = ["keccak"] }
-utils = { path = "../utils/" }
+utils = { path = "../utils/", default-features = false }
 
 # serialization
 serde_json = "1.0.48"
@@ -43,5 +43,5 @@ serde_json = "1.0.48"
 [features]
 default = ["parallel", "wasmer/sys-default"]
 fullmerkletree = []
-parallel = ["ark-ec/parallel", "ark-ff/parallel", "ark-std/parallel", "ark-groth16/parallel"]
+parallel = ["ark-ec/parallel", "ark-ff/parallel", "ark-std/parallel", "ark-groth16/parallel", "utils/parallel"]
 wasm = ["wasmer/js", "wasmer/std"]
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ark-ff = { version = "0.3.0", default-features = false, features = ["parallel", "asm"] }
+ark-ff = { version = "0.3.0", default-features = false, features = ["asm"] }
 num-bigint = { version = "0.4.3", default-features = false, features = ["rand"] }
 
 [dev-dependencies]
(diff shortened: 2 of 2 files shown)


rln/Cargo.toml
```diff
@@ -33,6 +33,7 @@ num-bigint = { version = "0.4.3", default-features = false, features = ["rand"]
 num-traits = "0.2.11"
 once_cell = "1.14.0"
 rand = "0.8"
+rand_chacha = "0.3.1"
```
fryorcraken, line 36: I am worried about the new dependency here. How does it affect the wasm blob size? I intended to add some size checking as part of the CI but it was not as straightforward as I hoped because cargo make is used instead of npm.

Cc @richard-ramos

rln/src/protocol.rs
```diff
@@ -338,13 +339,31 @@ pub fn compute_tree_root(
...
     let mut rng = thread_rng();
     let identity_secret = Fr::rand(&mut rng);
     let id_commitment = poseidon_hash(&[identity_secret]);
     (identity_secret, id_commitment)
 }
 
+// Generates a tupe (identity_secret, id_commitment) where
+// identity_secret is random and id_commitment = PoseidonHash(identity_secret)
+// RNG is instantiated using 20 rounds of ChaCha seeded with the hash of the input
+pub fn seeded_keygen(signal: &[u8]) -> (Fr, Fr) {
```
fryorcraken, line 353: this function can accept a 1 byte signal, which I assume, is not secure.

I'd suggest to fix the signal size (and I am guessing that it should be 32 bytes).

fryorcraken: Fair point.

rln/Cargo.toml
richard-ramos: Right now building this branch, i end up with 530KB which is way better than the 838KB of the version published https://unpkg.com/@waku/rln@0.0.8-964df01/bundle/assets/rln_wasm_bg-0185b546.wasm
```
-rw-rw-r-- 1 richard richard  17K sep 24 12:15 rln_wasm_bg.js
-rw-rw-r-- 1 richard richard 530K sep 30 08:42 rln_wasm_bg.wasm
-rw-rw-r-- 1 richard richard 1.1K sep 30 08:42 rln_wasm_bg.wasm.d.ts
-rw-rw-r-- 1 richard richard 3.2K sep 30 08:42 rln_wasm.d.ts
-rw-rw-r-- 1 richard richard  20K sep 30 08:42 rln_wasm.js
```

richard-ramos (approved):

rln/Cargo.toml
fryorcraken: What changed ???

fryorcraken:
Moved the discussion to https://github.com/status-im/nwaku/issues/1278 on whether we should use that for now.

PR #59, opened 2022-10-03
fix(rln): fix order of inputs in comments

s1fr0:
Fixes order of inputs to generate and verify procs.


richard-ramos (approved):
:heavy_check_mark:

PR #61, opened 2022-10-07
feat(RLN): Proof verification using provided roots

s1fr0:
This PR implements a new API called `verify_with_roots` which checks validity of RLN zk proof, similarly as `verify_rln_proof`. 

In contrast to the latter, this new API checks the proof's root value against a set of roots provided by the user, allowing to verify roots not necessarily against the current internal state of the RLN object. 

The input sets of roots can be empty: in this case, no check is done on proof's root value and the proof will be correctly verified if the other proof values are correct.


richard-ramos (approved):

rymnc (approved):
LGTM, thanks 😁

PR #73, opened 2022-11-15
fix(ci): exclude rln-wasm since it has its own step now

rymnc:
Exclude rln-wasm compilation in linux and macos steps


rymnc:
@s1fr0 please note that in [68bc1c4](https://github.com/vacp2p/zerokit/pull/73/commits/68bc1c4f7c9a67df8c3f84c0e2ae205bb40ec0d4) I removed the commit hash from the tarballs because this would not allow for a deterministic url for the nightly assets.

rymnc:
For actual releases, we can include the commit hash.

PR #84, opened 2022-12-06
refactor(rln): update APIs after circuit update

s1fr0:
This PR updates some zerokit RLN APIs in light of the security design changes discussed in https://github.com/vacp2p/research/issues/152, https://github.com/Rate-Limiting-Nullifier/rln_circuits/issues/1 and implemented in https://github.com/Rate-Limiting-Nullifier/rln_circuits/pull/2.

In particular external_nullifier is now computed as the Poseidon hash of epoch and RLN identifier, while in generating shares the nullifier is computed only by hashing the `a1` coefficient (see above issues for more details).

**Note that all circuit resources had to be re-generated and that new circuits are not compatible with previous ones.**

@richard-ramos @fryorcraken could you please ensure that these changes doesn't break anything in rln-wasm and js-rln ? (should not, but better we double check)

FYI: @curryrasul @AtHeartEngineer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
       - '!private-settlement/src/**'
       - '!rln-wasm/**'
       - '!rln/src/**'
+      - '!rln/resources/**'
       - '!semaphore/src/**'
       - '!utils/src/**'
   pull_request:
@@ -19,6 +20,7 @@ on:
       - '!private-settlement/src/**'
       - '!rln-wasm/**'
       - '!rln/src/**'
+      - '!rln/resources/**'
       - '!semaphore/src/**'
       - '!utils/src/**'
 


rymnc (approved):
LGTM! Thanks for the quick PR :)

fryorcraken:
@richard-ramos you ok to handle the query from @s1fr0 ?

oskarth:
Does this imply any spec changes? If so this needs a corresponding issue in RFC repo.

We need to have the RFC up to date and alive.

oskarth (changes requested):
See below

rymnc:
> See above

Tracking here https://github.com/vacp2p/rfc/issues/560

richard-ramos (approved):
It does not break any API but will require the following actions
In JS-Land:
1. Edit the Cargo.toml of rln-wasm to increase the version number
2. Publish a new version of @waku/zerokit-rln-wasm, with `cargo make publish`
3. Publish a new version of js-rln that uses the version of the package published in last step

In GO-Land:
1. Update [go-zerokit-rln](https://github.com/waku-org/go-zerokit-rln) so it uses latest zerokit version, and update all static libraries
2. Update go-waku to use latest go-zerokit-rln

rymnc:
@richard-ramos, please can you check if the nightly wasm artifact [here](https://github.com/vacp2p/zerokit/releases/tag/nightly) is suitable till we complete the steps you mentioned?

Having a new release is preferred since we are breaking APIs.

oskarth (approved):
Approving assuming we are tracking relevant changes in implementations too:

- Spec PR is done so that's good
- https://github.com/vacp2p/zerokit/pull/84#pullrequestreview-1208860932 this should have issues in nwaku/go-waku/js-waku so we make sure to update all implementations, or at least capture that this has to be done
- who will update nwaku?

rymnc:
We're tracking the nwaku update here - https://github.com/waku-org/nwaku/issues/1451. I can take it up after we're done with https://github.com/waku-org/nwaku/pull/1457

Issue #65, opened 2022-11-02
feat(utils): Merkle Tree persistence


s1fr0:
# Problem

Currently, the `merkle_tree` implementations in the `utils` module don't allow persistence of the created/processed Merkle tree, requiring applications to initialize and iteratively populate it for example, at each start-up.

We would like to extend/refactor/integrate a new implementation of Merkle tree that allows local persistence and the restore of the full state of a tree. 

Ideally, the new implementation should support fixed depth trees, implementable hash and databases and should not represent an evident regression in terms of performances and storage requirements with respect to the Merkle Tree implementation present in zerokit (in other words, a hashtable-based DB to store tree nodes is envisioned).

# Current development

The [PSE](https://github.com/privacy-scaling-explorations) team is already working on an implementation that could suit zerokit needs: the module is called [pmtree](https://github.com/Rate-Limiting-Nullifier/pmtree/) and EDIT: its first fully-featured implementation has been released.

## Integration in zerokit

At the moment zerokit requires basic tree APIs i.e., 
- (indexed) (batch) insert;
- deletion;
- proof creation/verification;
- tree root retrieval;
- Poseidon as main hash function;
- store tree state to file;
- restore tree state from file.
 
EDIT: These APIs have been implemented in pmtree.

Vac and PSE are in close collaboration in regards to the development of `pmtree` and its integration in zerokit would start as long as most of the above APIs are ready.

# Acceptance criteria
- [ ] Expand/refactor/integrate a new Merkle Tree implementation allowing persistence and restore of the tree full state with minor/none regression in performances and storage requirements.

PR #86, opened 2022-12-09
feat(rln): integrate pmtree for MT persistance

s1fr0:
This PR is a first step towards a fully working implementation of persisted Merkle trees in zerokit.

The PR integrates the pmtree module developed by PSE, as described in https://github.com/vacp2p/zerokit/issues/65, and checks correctness of implementations using SledDB for on-disk persistence and HashMap for in-memory storage.

A new feature pmtree is added in order to swap from the current default OptimalMerkleTree implementation to pmtree. However, at the moment this feature is disabled (will fallback to OptimalMerkleTree), since some pmtree's APIs still need to be aligned with zerokit module and other MT implementations.

Follow-up PRs will ensure that all 3 FullMerkleTree, OptimalMerkleTree and pmtree implementations are aligned and interchangeable by enabling the corresponding feature flag.

@curryrasul @AtHeartEngineer

### DIFF
--- a/rln/Cargo.toml
+++ b/rln/Cargo.toml
@@ -11,9 +11,6 @@ doctest = false
 
 
 [dependencies]
-pmtree = { git = "https://github.com/Rate-Limiting-Nullifier/pmtree" }
-sled = "0.34.7"
-
 # ZKP Generation
 ark-ec = { version = "0.3.0", default-features = false }
 ark-ff = { version = "0.3.0", default-features = false, features = [ "asm"] }
@@ -45,9 +42,15 @@ utils = { path = "../utils/", default-features = false }
 # serialization
 serde_json = "1.0.48"
 
+[dev-dependencies]
+pmtree = { git = "https://github.com/Rate-Limiting-Nullifier/pmtree" }
+sled = "0.34.7"
+
 [features]
 default = ["parallel", "wasmer/sys-default"]
 parallel = ["ark-ec/parallel", "ark-ff/parallel", "ark-std/parallel", "ark-groth16/parallel", "utils/parallel"]
 wasm = ["wasmer/js", "wasmer/std"]
 fullmerkletree = ["default"]
+
+# Note: pmtree feature is still experimental
 pmtree = ["default"]
--- a/rln/src/poseidon_tree.rs
+++ b/rln/src/poseidon_tree.rs
@@ -4,25 +4,13 @@
 
 use crate::circuit::Fr;
 use crate::poseidon_hash::poseidon_hash;
-use crate::utils::{bytes_le_to_fr, fr_to_bytes_le};
 use cfg_if::cfg_if;
-
-use pmtree::*;
-use sled::Db as Sled;
-use std::collections::HashMap;
-use std::fs;
-use std::path::Path;
 use utils::merkle_tree::*;
 
 // The zerokit RLN default Merkle tree implementation is the OptimalMerkleTree.
-// To switch to FullMerkleTree or PMTree implementations, it is enough to enable the fullmerkletree or pmtree feature, respectively
+// To switch to FullMerkleTree implementation, it is enough to enable the fullmerkletree feature
 
 cfg_if! {
-    // Not all used Merkle tree APIs are supported by pmtree. We comment for now.
-    // if #[cfg(feature = "pmtree")] {
-    //     pub type PoseidonTree = pmtree::MerkleTree<SledDB, PoseidonHash>;
-    //     pub type MerkleProof = pmtree::MerkleProof<PoseidonHash>;
-    // } else
     if #[cfg(feature = "fullmerkletree")] {
         pub type PoseidonTree = FullMerkleTree<PoseidonHash>;
         pub type MerkleProof = FullMerkleProof<PoseidonHash>;
@@ -49,101 +37,6 @@ impl utils::merkle_tree::Hasher for PoseidonHash {
     }
 }
 
-// The pmtree Hasher trait used by pmtree Merkle tree
-impl pmtree::Hasher for PoseidonHash {
-    type Fr = Fr;
-
-    fn default_leaf() -> Self::Fr {
-        Fr::from(0)
-    }
-
-    fn serialize(value: Self::Fr) -> Value {
-        fr_to_bytes_le(&value)
-    }
-
-    fn deserialize(value: Value) -> Self::Fr {
-        let (fr, _) = bytes_le_to_fr(&value);
-        fr
-    }
-
-    fn hash(inputs: &[Self::Fr]) -> Self::Fr {
-        poseidon_hash(inputs)
-    }
-}
-
-// pmtree supports in-memory and on-disk databases (Database trait) for storing the Merkle tree state
-
-// We implement Database for hashmaps, an in-memory database
-struct MemoryDB(HashMap<DBKey, Value>);
-
-impl Database for MemoryDB {
-    fn new(_dbpath: &str) -> Result<Self> {
-        Ok(MemoryDB(HashMap::new()))
-    }
-
-    fn load(_dbpath: &str) -> Result<Self> {
-        Err(Error("Cannot load in-memory DB".to_string()))
-    }
-
-    fn get(&self, key: DBKey) -> Result<Option<Value>> {
-        Ok(self.0.get(&key).cloned())
-    }
-
-    fn put(&mut self, key: DBKey, value: Value) -> Result<()> {
-        self.0.insert(key, value);
-
-        Ok(())
-    }
-}
-
-// We implement Database for sled DB, an on-disk database
-struct SledDB(Sled);
-
-impl Database for SledDB {
-    fn new(dbpath: &str) -> Result<Self> {
-        if Path::new(dbpath).exists() {
-            match fs::remove_dir_all(dbpath) {
-                Ok(x) => x,
-                Err(e) => return Err(Error(e.to_string())),
-            }
-        }
-
-        let db: Sled = match sled::open(dbpath) {
-            Ok(db) => db,
-            Err(e) => return Err(Error(e.to_string())),
-        };
-
-        Ok(SledDB(db))
-    }
-
-    fn load(dbpath: &str) -> Result<Self> {
-        let db: Sled = match sled::open(dbpath) {
-            Ok(db) => db,
-            Err(e) => return Err(Error(e.to_string())),
-        };
-
-        if !db.was_recovered() {
-            return Err(Error("Trying to load non-existing database!".to_string()));
-        }
-
-        Ok(SledDB(db))
-    }
-
-    fn get(&self, key: DBKey) -> Result<Option<Value>> {
-        match self.0.get(key) {
-            Ok(value) => Ok(value.map(|val| val.to_vec())),
-            Err(e) => Err(Error(e.to_string())),
-        }
-    }
-
-    fn put(&mut self, key: DBKey, value: Value) -> Result<()> {
-        match self.0.insert(key, value) {
-            Ok(_) => Ok(()),
-            Err(e) => Err(Error(e.to_string())),
-        }
-    }
-}
-
 ////////////////////////////////////////////////////////////
 /// Tests
 ////////////////////////////////////////////////////////////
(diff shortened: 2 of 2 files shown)


oskarth (commented):
We should make sure the DB persistence is compatible with e.g. nwaku and Waku Product needs. I flagged this a bunch of times with Sanaz and in #zerokit, but I'm not sure this conversation was actually had.

The concern is if we use different DBs how this impacts complexity, UX, migrations etc.

cc @LNSD @fryorcraken @jm-clius @richard-ramos

oskarth:
Also great work! Just need to make sure we think of compatibility and how this will work in production :)

rymnc:
> We should make sure the DB persistence is compatible with e.g. nwaku and Waku Product needs. I flagged this a bunch of times with Sanaz and in #zerokit, but I'm not sure this conversation was actually had.
> 
> The concern is if we use different DBs how this impacts complexity, UX, migrations etc.
> 
> cc @LNSD @fryorcraken @jm-clius @richard-ramos

imo if we use an embedded db it should not affect nwaku/other implementations since the persistence itself would be abstracted away. If we use an external db then these issues would arise. wdyt?

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+    }
+
+    fn serialize(value: Self::Fr) -> Value {
+        fr_to_bytes_le(&value)
+    }
+
+    fn deserialize(value: Value) -> Self::Fr {
+        let (fr, _) = bytes_le_to_fr(&value);
+        //let c = std::io::Cursor::new(value);
+        //Fr::deserialize(c).unwrap()
```
rymnc, line 67:
```suggestion
```

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+        Ok(SledDB(db))
+    }
+
+    fn get(&self, key: DBKey) -> Result<Option<Value>> {
+        Ok(self.0.get(key).unwrap().map(|val| val.to_vec()))
+    }
+
+    fn put(&mut self, key: DBKey, value: Value) -> Result<()> {
+        self.0.insert(key, value).unwrap();
+        self.0.flush().unwrap();
```
rymnc, line 129: Not sure if we should flush every time we set a value.
Maybe we should add a new function called `flush` to the `Database` trait, and have it do nothing for in-memory dbs. wdyt?

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+
+        Ok(SledDB(db))
+    }
+
+    fn get(&self, key: DBKey) -> Result<Option<Value>> {
+        Ok(self.0.get(key).unwrap().map(|val| val.to_vec()))
+    }
+
+    fn put(&mut self, key: DBKey, value: Value) -> Result<()> {
+        self.0.insert(key, value).unwrap();
```
rymnc, line 128: We should return the error here, not unwrap
```suggestion
        self.0.insert(key, value)?;
```
Or even better, have an enum for errors so that they can be appropriately handled by users.

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+
+        if !db.was_recovered() {
+            return Err(Error("Trying to load non-existing database!".to_string()));
+        }
+
+        Ok(SledDB(db))
+    }
+
+    fn get(&self, key: DBKey) -> Result<Option<Value>> {
+        Ok(self.0.get(key).unwrap().map(|val| val.to_vec()))
```
rymnc, line 124: Please return the error here

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+    fn new(dbpath: &str) -> Result<Self> {
+        if Path::new(dbpath).exists() {
+            fs::remove_dir_all(dbpath).unwrap();
+        }
+        let db = sled::open(dbpath).unwrap();
+        Ok(SledDB(db))
+    }
+
+    fn load(dbpath: &str) -> Result<Self> {
+        let db = sled::open(dbpath).unwrap();
```
rymnc, line 114:
```suggestion
        let db = sled::open(dbpath)?;
```

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+
+// We implement Database for sled DB, an on-disk database
+struct SledDB(Sled);
+
+impl Database for SledDB {
+    fn new(dbpath: &str) -> Result<Self> {
+        if Path::new(dbpath).exists() {
+            fs::remove_dir_all(dbpath).unwrap();
+        }
+        let db = sled::open(dbpath).unwrap();
```
rymnc, line 109:
```suggestion
        let db = sled::open(dbpath)?;
```

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+    }
+}
+
+// We implement Database for sled DB, an on-disk database
+struct SledDB(Sled);
+
+impl Database for SledDB {
+    fn new(dbpath: &str) -> Result<Self> {
+        if Path::new(dbpath).exists() {
+            fs::remove_dir_all(dbpath).unwrap();
```
rymnc, line 107:
```suggestion
            fs::remove_dir_all(dbpath)?;
```

rln/src/poseidon_tree.rs
```diff
@@ -105,5 +242,270 @@ mod test {
...
+        println!(
+            "   - Pmtree-HashMap Merkle Tree: {:?}",
+            Duration::from_nanos((upd_time_pm_memory / sample_size).try_into().unwrap())
+        );
+
+        println!(
+            "   - Pmtree-Sled Merkle Tree: {:?}",
+            Duration::from_nanos((upd_time_pm_sled / sample_size).try_into().unwrap())
+        );
+    }
```
rymnc, line 255: Maybe we can benchmark using the in-built lib - https://doc.rust-lang.org/1.16.0/book/benchmark-tests.html? (the lib is unstable, so up to you)

rln/src/poseidon_tree.rs
```diff
@@ -36,16 +49,101 @@ impl Hasher for PoseidonHash {
...
+}
+
+// pmtree supports in-memory and on-disk databases (Database trait) for storing the Merkle tree state
+
+// We implement Database for hashmaps, an in-memory database
+struct MemoryDB(HashMap<DBKey, Value>);
+
+impl Database for MemoryDB {
+    fn new(_dbpath: &str) -> Result<Self> {
```
rymnc, line 82: As we discussed, maybe each db that implements this trait should have a custom config in the `new` fn. This way we can set db-specific configuration as well.

rymnc (commented):
LGTM in general, just a few questions

rln/Cargo.toml
```diff
@@ -11,6 +11,8 @@ doctest = false
 
 
 [dependencies]
+pmtree = { git = "https://github.com/Rate-Limiting-Nullifier/pmtree" }
```
richard-ramos, line 14: What do you think of locking the pmtree version by using a commit, or maybe publishing `pmtree` so if a breaking change is introduced, we can still build zerokit? 
```suggestion
pmtree = { git = "https://github.com/Rate-Limiting-Nullifier/pmtree"  rev = "14d58fd" }
```

curryrasul: I think @s1fr0 just forgot to do it, It's always a good practice to lock on commits.

richard-ramos:
> Another relevant fact we're considering is that in WASM we cannot use straightforwardly common DBs implementations. Hence, we're want to reach enough flexibility on pmtree side first so that this can ultimately be possible (e.g. support IndexedDB or others WASM-friendly DBs).

I'm very curious in how we could persist data when compiling to WASM. Sounds very interesting! I imagine we could use `indexed_db_futures` or `idb` or local storage from `web_sys` and create a struct using their functions and implementing the Database trait, or maybe if we wanted to give more freedom to users to, for example, interact with some restful service for DB purposes, I imagine that we could add a function like this that would receive functions from javascript containing the implementation of `get`, and `put`

```rust
#[wasm_bindgen(js_name = setPMTreeDB)]
pub fn wasm_set_pmtree_db(getCB: &js_sys, putCB: &js_sys) {
   // TODO: pass an instance of `JSDatabase` built with these callbacks to pmtree somehow
   // 
}

impl Database for JSDatabase {
    fn new(_dbpath: &str) -> Result<Self> {
        Err(Error("Cannot create new JSDatabase".to_string()))
    }

    fn load(_dbpath: &str) -> Result<Self> {
        Err(Error("Cannot load JSDatabase".to_string()))
    }

    fn get(&self, key: DBKey) -> Result<Option<Value>> {
    	let this = JsValue::null();
    	let k = JsValue::from(key);
        let result = getCB.call1(&this, k);
        Ok(result))
    }

    fn put(&mut self, key: DBKey, value: Value) -> Result<()> {
        let this = JsValue::null();
        let k = JsValue::from(key);
        let v = JsValue::from(value);
        let result = putCB.call2(&this, k, v));
        Ok(())
    }
}
```

Then in javascript something like this could happen:

```js
// In javascript

function get(key){
   // TODO: get key somehow
   return key;
}

function put(key, value){
    // TODO: store key value
}

setPMTreeDB(get, put);
```
but this would assume that pmtree has a `pub fn default_with_db(depth: usize, db: &JSDatabase) -> Result<Self>` that would let us set a custom DB without pmtree calling `new` or `load`

rln/src/poseidon_tree.rs
curryrasul: That's what I was talking about before, but atm I don't know how to make simple & clear error handling. The problem is in `Error` compatibility: pmtree Database trait methods return custom result (defined in pmtree) and Sled::Db return their own errors.

Need to discuss and decide, maybe I can reimplement something in pmtree, so error handling will be better.

rln/src/poseidon_tree.rs
```diff
@@ -105,5 +242,270 @@ mod test {
...
+
+        let expected_identity_path_index: Vec<u8> =
+            vec![1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+
+        assert_eq!(path_elements, expected_path_elements);
+        assert_eq!(identity_path_index, expected_identity_path_index);
+
+        // We check correct verification of the proof
+        assert!(tree.verify(&id_commitment, &merkle_proof));
+
```
curryrasul, line 508: I'm not sure, but I think you need to add something like:
```rust
fs::remove_dir_all("pmtree_db").expect("Error removing db");
```
because this test creates repo/db. And if this test is run twice it'll give you errors, as db exists.

curryrasul:
Great work btw, thanks for fast integration!

rln/src/poseidon_tree.rs
curryrasul: Hm, but then what if user accidentally calls new with existing path?

Did you consider making it as I did in tests:
```rust
    fn new(dbpath: &str) -> Result<Self> {
        let db = sled::open(dbpath).unwrap();
        if db.was_recovered() {
            return Err(Error("Database exists, try load()!".to_string()));
        }

        Ok(MySled(db))
    }

    fn load(dbpath: &str) -> Result<Self> {
        let db = sled::open(dbpath).unwrap();

        if !db.was_recovered() {
            fs::remove_dir_all(dbpath).expect("Error removing db");
            return Err(Error("Trying to load non-existing database!".to_string()));
        }

        Ok(MySled(db))
    }
```

rln/src/poseidon_tree.rs
rymnc: How about a generic?

curryrasul: How do you see it ? @rymnc

rymnc: Maybe something like this in pmtree -

(associated type method) -
```rust
trait Database {
  type DbConfig;
  
  fn new(path: &str, config: Self::DbConfig) -> Result<Self>;
  // other funcs
}
```

and then this in rln -
```rust
impl Database for MemoryDB {
  type DbConfig = MemoryDbConfig;
  
  fn new(_dbpath: &str, config: MemoryDbConfig) -> Result<Self> {
    // fn
  }
  
  // other funcs
}
```

OR

(constraints have to be the same across all configs) (generally not used)
(generic trait method) -
in pmtree -
```rust
trait Database<T> {
  fn new(path: &str, config: T) -> Result<Self>;
  // other funcs
}
```

in rln -
```rust
impl Database<MemoryDbConfig> for MemoryDB {
  fn new(_dbpath: &str, config: MemoryDbConfig) -> Result<Self> {
    // fn
  }
  
  // other funcs
}
```

It doesn't make a lot of sense choosing the latter option, because any db would have a concrete type for their config. Let's go with the first option?

curryrasul: But what this Config specifies? How it should affect tree implementation, traits, etc. ?

rymnc: This config is db specific. For example, sled db has a flush rate of 500ms. If we wanted to configure it, we can do so here, and by exposing this option in the `new` fn, we can allow users of this lib to set it to whichever value they want. This is just an example, and can evolve with different dbs we use, and how we want to allow the users to configure them.

curryrasul: Oh, yes, I got it! That would be good feature, I can add it:)

rymnc: Thank you!😁

rln/src/poseidon_tree.rs
rymnc: Thanks!

rln/src/poseidon_tree.rs
rymnc: Absolutely, let's go with the first method :)

rln/src/poseidon_tree.rs
rymnc: Thanks!

rln/src/poseidon_tree.rs
rymnc: Thanks!

rln/src/poseidon_tree.rs
rymnc: Thanks!

rln/src/poseidon_tree.rs
rymnc: Sounds good. Tracking here - https://github.com/vacp2p/zerokit/issues/89

rymnc (approved):
LGTM, we can integrate the `DbConfig` in a new PR :)

rln/src/poseidon_tree.rs
danielSanchezQ: Probably late to the dance. But unless targeting no-std, you could probably use thiserror crate instead.

Issue #97, opened 2023-01-16
Check and add licenses


s1fr0:
We need to add a license to zerokit, likely MIT and/or Apache v2.

Before doing so, we need to check and ensure that all dependencies and integrated code snippets are covered by a license compatible with the one chosen.

# Acceptance criteria
- [ ] Ensure dependencies and integrated code all have a compatible license
- [ ] Add a compatible license to zerokit

PR #119, opened 2023-02-24
chore: Add MIT/Apache dual license

oskarth:
Addresses https://github.com/vacp2p/zerokit/issues/97

I should've done this at the very beginning, but forgot to.

MIT/Apache is the standard license used in Rust ecosystem, and there's no reason to do anything else since we aren't doing or using anything weird. This is also evident in `cargo-license` printouts.

See equivalent on ark-circom side https://github.com/gakonst/ark-circom/commit/8f6fcaf40b182e745126a9ae07a653f0ef97df9a

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,203 @@
+Copyright (c) 2022 Vac Research
+
+                              Apache License
+                        Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


rymnc (approved):
LGTM, do we need someone from legal to review it too?

PR #120, opened 2023-02-24
Update documentation for building and testing

### LINKED ISSUE #106: Update docs on how to build/test submodules
See [#101 comment](https://github.com/vacp2p/zerokit/pull/101#issuecomment-1400087974)

rahulghangas:
As title

- [x] closes #106

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,10 @@
 
 all: .pre-build build
 
-.pre-build:
+.fetch-submodules:
+	@git submodule update --init --recursive
+
+.pre-build: .fetch-submodules
 ifeq (, $(shell which cargo-make))
 	@cargo install --force cargo-make
 endif


Makefile
```diff
@@ -1,13 +1,17 @@
-.PHONY: all test clean
+.PHONY: all installdeps build test clean
 
-all: .pre-build
-	@cargo make build
+all: .pre-build build
 
 .pre-build:
```
rymnc, line 5: Should this include installing the submodules too?

If so, we can update the ci workflows too

rymnc: Awesome, I meant that ci should use the make target (which includes installing submodules), instead of installing submodules on its own. Can be addressed in the future though :)

rymnc (approved):
LGTM, the ci improvements discussed can be addressed in a follow up pr

PR #127, opened 2023-03-03
chore(rln): ensure all dependencies have fixed revision 

### LINKED ISSUE #126: chore(rln): ensure all dependencies have fixed revision
Ensure all dependencies have fixed revision in order to publish a stable version of rln

tyshko-rostyslav:
Closes #126

### DIFF
--- a/rln/Cargo.toml
+++ b/rln/Cargo.toml
@@ -13,39 +13,38 @@ doctest = false
 
 [dependencies]
 # ZKP Generation
-ark-ec = { version = "0.3.0", default-features = false }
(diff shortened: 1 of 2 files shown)


rymnc:
Shouldn't we do this for the other packages too? (rln-wasm)

tyshko-rostyslav:
> Shouldn't we do this for the other packages too? (rln-wasm)

I updated all connected crates, but `rln-wasm` is actually independent, so I guess not.

rymnc:
Just checked out the issue, looks like it's specific to rln. LGTM then!

rymnc (approved):

PR #130, opened 2023-03-20
fix(semaphore): use fixed rev

rymnc:
Temporary fix for failing CI ref: https://github.com/vacp2p/zerokit/actions/runs/4463841234/jobs/7839439039
ref issue: https://github.com/vacp2p/zerokit/issues/131


tyshko-rostyslav:
I'm not completely sure we should do anything with `semaphore` now. When I talked about it with Giuseppe, when he was working on it, he said not to do anything, as `semaphore` had to be rewritten completely.

rymnc:
It's just dependency freezing at the moment. the changes he had in mind don't relate to the ark-circom dependency

Issue #135, opened 2023-03-23
add `wasm-strip` to reduce size even more 


tyshko-rostyslav:
Using `wasm-strip` on the latest master made `rln_wasm_bg.wasm` file to be of size 509KB, instead of 612KB

PR #137, opened 2023-03-23
add wasm-strip to reduce size even more 

tyshko-rostyslav:
Closes: #135

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ ifdef CI
 endif
 
 installdeps: .pre-build
+	@git clone --recursive https://github.com/WebAssembly/wabt.git
+	@cd wabt && make
+	@sudo make install
 
 build: .pre-build
 	@cargo make build
--- a/Makefile
+++ b/Makefile
@@ -12,9 +12,10 @@ ifdef CI
 endif
 
 installdeps: .pre-build
+	@sudo apt-get update
+	@sudo apt-get install -y cmake ninja-build
 	@git clone --recursive https://github.com/WebAssembly/wabt.git
-	@cd wabt && make
-	@sudo make install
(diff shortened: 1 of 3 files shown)


rymnc:
Should wasm-strip be added to the deps in the Makefile?
You can add the post-build task to the dependency array https://github.com/vacp2p/zerokit/blob/658d07ae8cd7148a7d7a8237d2400bd88c7a7abf/rln-wasm/Makefile.toml#L8-L14 so that it runs automatically

tyshko-rostyslav:
> Should wasm-strip be added to the deps in the Makefile? You can add the post-build task to the dependency array
> 
> https://github.com/vacp2p/zerokit/blob/658d07ae8cd7148a7d7a8237d2400bd88c7a7abf/rln-wasm/Makefile.toml#L8-L14
> 
> so that it runs automatically

wasm-strip, cannot be a dependency of `build`, since it needs a built `.wasm` file to work with first.

rymnc:
pack-rename runs after the file is built, isn't it?

tyshko-rostyslav:
> The error in CI is because wasm-strip isn't installed, please install it as a dependency in the root Makefile

I think it cannot find the file

rymnc:
With this -
```
[tasks.post-build]
script = "wasm-strip pkg/rln_wasm_bg.wasm"
```
This is the build error -
```
[cargo-make] INFO - Running Task: pack-rename
[cargo-make] INFO - Running Task: post-build
/var/folders/v_/4f04x6_901qckppx8f0lj_h00000gn/T/fsio_E3zqzGVGzp.sh: line 3: wasm-strip: command not found
[cargo-make] ERROR - Error while executing command, exit code: 127
```

tyshko-rostyslav:
> The "file" it cannot find is actually the executable for wasm-strip

Fixed

PR #138, opened 2023-03-23
fix(rln): move std::path to cfg_if block

rymnc:
std::path::Path is unused in the wasm context, therefore, moved the import to the cfg_if block


tyshko-rostyslav (approved):
LGTM

PR #140, opened 2023-03-28
Macro to call functions with an error message

tyshko-rostyslav:


rln-wasm/src/lib.rs
```diff
@@ -20,6 +20,64 @@ pub struct RLNWrapper {
...
+            let new_instance: &mut RLNWrapper = $instance.process();
+            if new_instance.instance.$method($($arg.process()),*).is_ok() {
+                Ok(())
+            } else {
+                Err($error_msg)
+            }
+
+        }
+    }
+}
```
rymnc, line 39: if the method returns a result, we can bubble the error up. please combine the errors so that it is more meaningful

tyshko-rostyslav: Methods that can be used here return `bool` only, no `Result`s

rymnc: we use .is_ok() though, am i missing something?

tyshko-rostyslav: I was wrong, pointing to a function with the same name but different signature.

Fixed as requested

rymnc: Awesome, thanks!

rymnc (approved):
LGTM, as a side note, whenever updating rln-wasm, please add @richard-ramos as a reviewer too :)

richard-ramos (approved):
I like this PR! Code is more elegant and cleaner. :100:

PR #141, opened 2023-03-29
Macro to call functions with an error message with output

tyshko-rostyslav:
Another variation of our call, this time when output is used


rln-wasm/src/lib.rs
```diff
@@ -20,6 +22,45 @@ pub struct RLNWrapper {
...
+        {
+            let mut output_data: Vec<u8> = Vec::new();
+            let new_instance = $instance.process();
+            if new_instance.instance.$method($($arg.process()),*, &mut output_data).is_ok() {
+                let result = Uint8Array::from(&output_data[..]);
+                std::mem::forget(output_data);
+                Ok(result)
+            } else {
+                std::mem::forget(output_data);
+                Err($error_msg.into())
```
rymnc, line 58: like the pr for rln-wasm, can we include the error messages here too?

tyshko-rostyslav: Fixed here: https://github.com/vacp2p/zerokit/pull/141/commits/ce5248aa4855b2f3e2228ca1b46051db33af1915

rymnc (approved):
LGTM, thanks for this.

Issue #139, opened 2023-03-28
feat(rln): Pmtree implementation


rymnc:
Currently, we have implemented `OptimalMerkleTree`(https://github.com/vacp2p/zerokit/blob/1f983bb23226e90b395ff58214f47ae3dca0d634/utils/src/merkle_tree/merkle_tree.rs#L69-L291), and `FullMerkleTree`(https://github.com/vacp2p/zerokit/blob/1f983bb23226e90b395ff58214f47ae3dca0d634/utils/src/merkle_tree/merkle_tree.rs#L297-L564). We need to do the following to ensure compatibility with the implementation of `Pmtree` similarly

- [x] Create a trait `ZerokitMerkleTree` or equivalent, declare the appropriate fns from FullMerkleTree and OptimalMerkleTree
  - [x] https://github.com/vacp2p/zerokit/pull/142
- [x] Update implementations of FullMerkleTree and OptimalMerkleTree to implement the ZerokitMerkleTree trait
  - [x] https://github.com/vacp2p/zerokit/pull/142
- [ ] Implement the ZerokitMerkleTree for a struct, PmTree (SledDB, for now), using the pmtree library
- [ ] Remove the tests of pmtree [here](https://github.com/vacp2p/zerokit/blob/1f983bb23226e90b395ff58214f47ae3dca0d634/rln/tests/poseidon_tree.rs#L85-L618) and make tests similar to the other 2 trees, flagged by the pmtree-ft feature
- [ ] Update the type of the [PoseidonTree](https://github.com/vacp2p/zerokit/blob/1f983bb23226e90b395ff58214f47ae3dca0d634/rln/src/poseidon_tree.rs#L18-L26) to use pmtree if the feature is enabled

This will allow downstream users to use pmtree for atomicity of insertions/deletions.

PR #142, opened 2023-03-30
feat: abstract shared behaviour into ZerokitMerkleTree

rymnc:
Creates 2 traits -
- `ZerokitMerkleTree`
- `ZerokitMerkleProof`

which, are implemented for `OptimalMerkleTree`, `FullMerkleTree`, `OptimalMerkleProof` and `FullMerkleProof` respectively

Addresses part of https://github.com/vacp2p/zerokit/issues/139


utils/src/merkle_tree/full_merkle_tree.rs
rymnc, line 1: note: this pr does not change any functionality of the tree, just split into its own file, and makes use of the trait defined in merkle_tree.rs

utils/src/merkle_tree/optimal_merkle_tree.rs
rymnc, line 1: note: this pr does not change any functionality of the tree, just split into its own file, and makes use of the trait defined in merkle_tree.rs

PR #147, opened 2023-04-24
Integrate pmtree into rln

rymnc:
Fixes https://github.com/vacp2p/zerokit/issues/139

Integrates pmtree into rln and uses it when `pmtree-ft` is enabled

- fix: use sled_adapter, rexports
- fix: remove dep on pmtree, use same as utils
- fix: use custom errors
- feat: added a `Config` associated type on `ZerokitMerkleTree`
- chore: `ZerokitMerkleTree::new` and `ZerokitMerkleTree::default` return Results now 

Next follow up PR:
- [ ] use pmtree by default
- [ ] allow passing in a config in the ffi


tyshko-rostyslav (approved):
LGTM

PR #153, opened 2023-05-02
fix: u64 to usize conversion

richard-ramos:
`usize::from_le_bytes` will take an array of length 2, 4 or 8 bytes depending on the target pointer size. Since wasm is a 32b platform, it will take 4 bytes instead of the 8 bytes defined by the protocol, so there was a failure while reading serialized values due to the number of bytes read being incorrect. I also update version rln-wasm to 0.0.9 (npm package needs to be updated to include this fix)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2405,7 +2405,7 @@ dependencies = [
 
 [[package]]
 name = "rln-wasm"
-version = "0.0.7"
+version = "0.0.9"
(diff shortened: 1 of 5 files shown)


tyshko-rostyslav (approved):
LGTM!

rymnc (approved):
LGTM, thanks for taking the time to run the integration tests with js

PR #154, opened 2023-05-03
fix: `len().to_le_bytes().to_vec()` contains 4 bytes on 32b and should contain 8

richard-ramos:
In https://github.com/vacp2p/zerokit/pull/153 I forgot to commit these changes that also deal with usize length :(
Sorry


rymnc (approved):
LGTM, thanks

PR #160, opened 2023-05-05
combined batch operations (insert + delete)

rymnc:
Allows insertions and deletions in the same batch operation.

Note: deletion is just zeroing out a given index

- feat: batch ops in ZerokitMerkleTree

Follow up PR: 
- [ ] public, ffi api and tests using them


rymnc:
update: there seems to be a bug with the implementations in OMT and FMT, fixing

tyshko-rostyslav (approved):
LGTM

PR #165, opened 2023-05-17
fix: wasm tests and ci

rahulghangas:
Unwrap result values in `rln-wasm` tests due to changed function signature
Fix CI, particularly `rl-wasm` tests were not being run

- [x] local testing


tyshko-rostyslav (approved):
LGTM

rymnc (approved):
LGTM, thanks for catching this!

Issue #89, opened 2022-12-12
Benchmarking using the inbuilt lib


rymnc:
As per the discussion here - https://github.com/vacp2p/zerokit/pull/86#discussion_r1044331978. When the benchmarking needs become more significant, we will use the inbuilt library or an alternative to benchmark to understand where the code can be optimized.

rymnc:
Moving forward with [criterion.rs](https://github.com/bheisler/criterion.rs)

PR #166, opened 2023-05-23
feat(utils): initialize benchmarks

rymnc:
- feat(utils): initialize benchmarks
- docs(utils): benchmark instructions

This PR solves a part of #89

example of the report generated -
image


rahulghangas (approved):
Good stuff!

PR #169, opened 2023-05-29
feat(rln): pmtree benchmarks

rymnc:
Closes https://github.com/vacp2p/zerokit/issues/89

image

pmtree using sled vs in-memory omt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2678,6 +2678,7 @@ dependencies = [
  "ark-std 0.4.0",
  "cfg-if",
  "color-eyre 0.6.2",
+ "criterion 0.4.0",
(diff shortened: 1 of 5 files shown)


rymnc:
Note to reviewers: 
the graph on the left is pmtree using sled db (fs db) vs optimal merkle tree using an in-memory db, performance is as expected.

We will tune sled db's configs to our use-case and create a follow up pr for it, as mentioned in https://github.com/vacp2p/zerokit/issues/151

curryrasul (approved):
LGTM! Thanks:)

tyshko-rostyslav (approved):
LGTM!

PR #171, opened 2023-05-29
feat(ci): compare benchmarks with master 

rymnc:
Compares and creates a comment with the benchmarks of each pr to avoid regressions


curryrasul (approved):
Great!

PR #173, opened 2023-05-29
fix(ci): run benchmarks only in PRs

rymnc:
Benchmarks should only run in PRs so they have a base branch to compare


rahulghangas (approved):
utAck

PR #178, opened 2023-06-15
feat(rln): expose set_metadata and get_metadata public and ffi apis

rymnc:
This feature is required so that downstream users of this library can set metadata related to the usage of the tree in the underlying database. This reduces client side complexity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2387,7 +2387,7 @@ dependencies = [
 [[package]]
 name = "pmtree"
 version = "1.0.0"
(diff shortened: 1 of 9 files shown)


richard-ramos (approved):
Cool PR. It's missing exposing the function in wasm  but that can be done in a separate PR

rymnc:
> Cool PR. It's missing exposing the function in wasm but that can be done in a separate PR

Good point! Will create a follow up pr for it.

PR #180, opened 2023-06-16
fix(rln): tree_config parsing

rymnc:
This PR resolves an error in which downstream libs were not able to pass a custom tree_config, this was due to using `as_str`, instead of 
`to_string`, which resulted in the default config to be used.


tyshko-rostyslav (approved):
Thanks! LGTM

PR #194, opened 2023-07-31
fix(crates): version tags

rymnc:
@tyshko-rostyslav, could you update the release on crates.io with these tags?


tyshko-rostyslav (approved):
LGTM

PR #195, opened 2023-07-31
fix(rln): atomic operation edge case

rymnc:
Tracking from https://github.com/waku-org/go-zerokit-rln/pull/12

This PR improves readability of the `override_range` fn, by matching on 4 input cases, and handles all of them 
appropriately.

### DIFF
--- a/rln/src/public.rs
+++ b/rln/src/public.rs
@@ -362,7 +362,7 @@ impl RLN<'_> {
         // We set the leaves
         self.tree
             .override_range(index, leaves, indices)
-            .map_err(|_| Report::msg("Could not perform the batch operation"))?;
+            .map_err(|e| Report::msg(format!("Could not perform the batch operation: {e}")))?;
         Ok(())
     }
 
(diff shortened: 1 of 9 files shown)


rln/src/public.rs
```diff
@@ -1476,6 +1478,66 @@ mod test {
...
+        let (root_after_deletion, _) = bytes_le_to_fr(&buffer.into_inner());
+
+        assert_ne!(root_after_insertion, root_after_deletion);
+
+        // We get the leaf
+        let mut output_buffer = Cursor::new(Vec::<u8>::new());
+        rln.get_leaf(set_index, &mut output_buffer).unwrap();
+        let (received_leaf, _) = bytes_le_to_fr(output_buffer.into_inner().as_ref());
+
+        assert_eq!(received_leaf, Fr::from(0));
```
rymnc, line 1546: this should address the edge case?

rymnc:
Finally resolved with @richard-ramos! 🚀 thanks for all the help debugging!

PR #199, opened 2023-08-04
chore: accept tree_config in new_with_params

richard-ramos:
go-waku embeds the zkey, vk and circuit in the binary, so it's not necessary to rely on external files when deploying go-waku service node. It uses `new_with_params` to be able to pass these parameters. 
However, this function does not allow passing the merkle tree configuration. 
This PR adds a new parameter `tree_config`  which would allow passing this configuration. 

In a future PR, rln-wasm should be updated to receive a JSON (not a string containing a json) in the `wasm_new` function

### DIFF
--- a/rln/src/public.rs
+++ b/rln/src/public.rs
@@ -117,6 +117,7 @@ impl RLN<'_> {
     /// - `circom_vec`: a byte vector containing the ZK circuit (`rln.wasm`) as binary file
     /// - `zkey_vec`: a byte vector containing to the proving key (`rln_final.zkey`) as binary file
     /// - `vk_vec`: a byte vector containing to the verification key (`verification_key.json`) as binary file
+    /// - `tree_config`: a reader for a string containing a json with the merkle tree configuration
     ///
     /// Example:
     /// ```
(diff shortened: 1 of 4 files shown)


rln/src/public.rs
```diff
@@ -134,18 +135,60 @@ impl RLN<'_> {
...
+            proving_key,
+            verification_key,
+            tree,
+        })
+    }
+
+    #[cfg(target_arch = "wasm32")]
     pub fn new_with_params(
         tree_height: usize,
-        #[cfg(not(target_arch = "wasm32"))] circom_vec: Vec<u8>,
```
rymnc, line 148: why is this line removed?

rln/src/public.rs
```diff
@@ -159,12 +202,9 @@ impl RLN<'_> {
         let tree = PoseidonTree::default(tree_height)?;
 
         Ok(RLN {
-            #[cfg(not(target_arch = "wasm32"))]
-            witness_calculator,
             proving_key,
             verification_key,
             tree,
-            #[cfg(target_arch = "wasm32")]
```
rymnc, line 167: why is this line removed?

rymnc (commented):
LGTM in general, but left a couple comments! thanks for this pr!

rln/src/public.rs
richard-ramos: The reason why some lines are removed is that there are now 2 `new_with_params` functions (in line 151 and 190). One of these drops the code/params that are not needed when compiling to wasm, and another one which is used for all other architectures.

I tried to keep a single new_with_params function but the code ended up being harder to read, so I ended up creating two, which simplifies reading the code with the tradeoff of adding (some) duplication of code

rymnc (approved):
LGTM, thanks for this!

PR #200, opened 2023-08-08
feat: print to stderr any error obtained while executing functions via FFI

richard-ramos:
This will help users of the library to get a reason on why a function is failing so we can help with support and reduce guesswork.


rymnc (approved):
LGTM as an initial iteration for error handling

tyshko-rostyslav (approved):
LGTM for now, we'll need to enhance it

PR #202, opened 2023-08-10
feat(rln-v2): update serde for rln-v2

rymnc:
In https://github.com/Rate-Limiting-Nullifier/circom-rln/blob/17f0fed7d8d19e8b127fd0b3e5295a4831193a0d/circuits/rln.circom#L8-L16 we see that there are 2 additional 
private inputs, `user_message_limit` and `message_id`, this PR adds it to the serde api's.


rymnc:
Note: the tests will fail till we get the new artifacts from the rln-v2 ceremony

richard-ramos (approved):

tyshko-rostyslav:
> Note: the tests will fail till we get the new artifacts from the rln-v2 ceremony

While it's not the case, maybe we should keep the failing code at a temporary branch?

rymnc:
> > Note: the tests will fail till we get the new artifacts from the rln-v2 ceremony
> 
> While it's not the case, maybe we should keep the failing code at a temporary branch?

Yes, merging into `rln-v2` branch

PR #206, opened 2023-08-18
chore(rln): expose seq_atomic_operation api

rymnc:
Exposes a `seq_atomic_operation` api that uses the `next_index` under the hood to sequentially write to the tree


tyshko-rostyslav (approved):
LGTM

PR #208, opened 2023-08-24
fix(rln): use config.open instead of sled::open

rymnc:
By using sled::open, we default to sled db's default config, which prevents us from opening the tree, thereby recreating it


rymnc:
Closing in favor of #209

PR #210, opened 2023-08-28
`message_id` range check

tyshko-rostyslav:
- Adds a range check before the proof generation to ensure the provided `messageId` is within the `userMessageLimit`
- Changed corresponding types to `Result`, as now they can emit errors
- Added function level documenation


rymnc (approved):
LGTM, sorry, thought i approved it last week!

PR #211, opened 2023-09-04
feat(rln-v2): new artifacts, removed older ones

rymnc:
Fetched from https://ceremony.pse.dev/projects/RLN%20Trusted%20Setup%20Ceremony
- wasm blobs generated using https://github.com/rate-limiting-nullifier/circom-rln
- verification keys generated from the zkeys

we also removed the rln circuits submodule


rymnc:
The only test failing is `compute_id_secret`, which is being taken care of in #217

PR #212, opened 2023-09-04
Fix tests

tyshko-rostyslav:


rymnc (approved):
LGTM since all unwrap's are within the test module

PR #217, opened 2023-09-18
Fix message

tyshko-rostyslav:
Fixing `test_recover_id_secret` test though fixing serialization


rln/src/protocol.rs
```diff
@@ -49,6 +49,7 @@ pub struct RLNProofValues {
     // Public Inputs:
     pub x: Fr,
     pub external_nullifier: Fr,
+    pub message_id: Fr,
```
rymnc, line 52: isn't message_id a private input?
https://github.com/Rate-Limiting-Nullifier/circom-rln/blob/b40dfa63b7b1248527d7ab417d0d9cf538cad93a/circuits/rln.circom#L40

rymnc (commented):
LGTM in general, curious about usage of `message_id`

rln/src/protocol.rs
tyshko-rostyslav: Added corresponding comment

rymnc: the struct RLNProofValues should only contain the data required to verify a proof, right? can we remove message_id from this?

rln/src/protocol.rs
```diff
@@ -382,13 +386,17 @@ pub fn deserialize_proof_values(serialized: &[u8]) -> (RLNProofValues, usize) {
...
+    all_read += read;
+
     (
         RLNProofValues {
             y,
             nullifier,
             root,
             x,
             external_nullifier,
+            message_id,
```
rymnc, line 399: this should be removed, right?

tyshko-rostyslav: Removed

rln/src/protocol.rs
tyshko-rostyslav: Removed, as requested

PR #226, opened 2023-12-07
chore: start optimizing wasm blob

rymnc:
Work with @weboko to optimize wasm usage in js-waku impls


seemenkina:
Done in PR #296

PR #227, opened 2024-01-10
Expose generate proof with witness ffi

alrevuelta:
* Expose generate proof with witness ffi. This aims to allow other programming languages using the ffi to provide its own witness.
* Based on v0.3.4 so that its compatible with RLN v1. Ignore diff since the PR can't be based in a tag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - master
+      - v0.3.5
     paths-ignore:
       - '**.md'
       - '!.github/workflows/*.yml'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,7 @@
 on:
   push:
     branches:
-      - v0.3.5
-      - add-witness
+      - master
     paths-ignore:
       - '**.md'
       - '!.github/workflows/*.yml'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - v0.3.5
+      - add-witness
     paths-ignore:
       - '**.md'
       - '!.github/workflows/*.yml'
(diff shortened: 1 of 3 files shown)


rln/src/public.rs
```diff
@@ -728,9 +728,9 @@ impl RLN<'_> {
         mut output_data: W,
     ) -> Result<()> {
         // We read input RLN witness and we serialize_compressed it
-        let mut witness_byte: Vec<u8> = Vec::new();
-        input_data.read_to_end(&mut witness_byte)?;
-        let (rln_witness, _) = proof_inputs_to_rln_witness(&mut self.tree, &witness_byte)?;
+        let mut input_byte: Vec<u8> = Vec::new();
```
alrevuelta, line 731: renamed since afaik the input here is not the witness

rymnc (approved):
LGTM in general, if we can fix the clippy issues it can be merged
if you want to test with rln-v1 you will have to open a pr to the v0.3.4 branch :)

rymnc:
> > if you want to test with rln-v1 you will have to open a pr to the v0.3.4 branch :)
> 
> Does this mean that latest master of zerokit doesn't work with rln v1?

yes because we do not maintain them both, after v0.3.4 (last release of rln-v1) we released v0.4.x for rln-v2

alrevuelta:
Okay, "based" now on v0.3.4. However since v0.3.4 is a tag, not a branch I can't really merge it there. If we wanted to merge this PR so that it can be used with rln v1, we would need to merge it somewhere else and then "retag" +1 v0.3.**5**.

Since I'm doing some tests/PoC by now, I will leave the PR by now without the intention of merging. Want to ensure first that it works end to end with eg go-rln.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - master
+      - v0.3.5
```
alrevuelta, line 4: To be removed, just forcing CI to be run, since this PR is not to be merged in master

alrevuelta:
* Skipping some CI checks, as they are related to linting problems already existing in master.
* Note that this PR is not merged in master.
* This PR is merged into a newly created branch `v0.3.5` that contains release `v0.3.4` changes plus this PR once merged.
* This new branch `v0.3.5` will be used to create a new release based on `v0.3.4` plus these changes.

PR #228, opened 2024-01-23
fix: add support to aarch64-linux-android

richard-ramos:
Otherwise, it will complain about -lunwind not found due to reasons described in https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -30,3 +30,6 @@ image = "ghcr.io/cross-rs/mips64el-unknown-linux-gnuabi64:latest"
 
 [target.mipsel-unknown-linux-gnu]
 image = "ghcr.io/cross-rs/mipsel-unknown-linux-gnu:latest"
+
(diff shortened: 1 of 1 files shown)


alrevuelta (approved):

rymnc:
this will go into the rln-v2 release process, is that okay?

rymnc:
err, please wait for checks to pass :)

richard-ramos:
oops, sorry!
I imagine the checks will pass since i modified Cross.toml only and that's not used in this project, but on go-zerokit-rln-*

richard-ramos:
> this will go into the rln-v2 release process, is that okay?

I also pushed it in https://github.com/vacp2p/zerokit/tree/v0.3.5-release commit history!

richard-ramos:
Just learned that cross is used here too :)

PR #233, opened 2024-03-27
feat: add functions for calculating witness that accepts merkle proof

adklempner:


rln/src/protocol.rs
```diff
@@ -350,6 +397,41 @@ pub fn proof_values_from_witness(rln_witness: &RLNWitnessInput) -> Result<RLNPro
     })
 }
 
+pub fn proof_values_from_witness_with_merkle_proof(
```
alrevuelta, line 400: Out of curiosity, is this needed for jswaku? I assume it only uses jswaku can only access what is exposed in rln-wasm?

alrevuelta (commented):
Thanks! Some comments:
* afaik `v0.4.x` only supports RLNv2 and `v0.3.x` is for RLNv1. Since The Waku Network (and Waku in general) is not yet using RLNv2, I would say we need to use the latest `v0.3.6` (RLN v1) as the baseline. I assume in js-waku you still point to that one?
* Perhaps`wasm_get_serialized_rln_witness_with_proof` is not needed?. I mean, you can generate the witness yourself without using zerokit (example using [go](https://github.com/waku-org/go-zerokit-rln/blob/master/rln/utils.go#L13-L26)) and then directly use `generate_rln_proof_with_witness` with your witness constructed in js-waku. Nothing against exposing it though, but feels more complex (requires modifying zerokit, extra call to wasm, some ugly serde, etc)
* So unless I'm missing something, I would say there is no need to modify zerokit for the PoC you are developing in js-waku. One interesting abstraction could be to have a `generate_rln_proof_with_merkle_proof` that gets everything: merkle proof, message hash, epoch, secret, etc. So instead of requiring to i) generate witness and ii) generate rln proof with witness, you can do everything in just 1 function call. But again, a mere nice to have.

In case that helps, [here is](https://github.com/waku-org/go-zerokit-rln/blob/14960f3aff2a50088f780c233bda492ddbe26162/rln/rln_test.go#L372-L392) how I did it in go.

rln/src/protocol.rs
adklempner: It's not needed, I forgot to remove this after realizing I only needed a new function to generate the witness

adklempner:
> * afaik `v0.4.x` only supports RLNv2 and `v0.3.x` is for RLNv1. Since The Waku Network (and Waku in general) is not yet using RLNv2, I would say we need to use the latest `v0.3.6` (RLN v1) as the baseline. I assume in js-waku you still point to that one?

Yes, this was my mistake.

> * Perhaps`wasm_get_serialized_rln_witness_with_proof` is not needed?. I mean, you can generate the witness yourself without using zerokit (example using [go](https://github.com/waku-org/go-zerokit-rln/blob/master/rln/utils.go#L13-L26)) and then directly use `generate_rln_proof_with_witness` with your witness constructed in js-waku. Nothing against exposing it though, but feels more complex (requires modifying zerokit, extra call to wasm, some ugly serde, etc)

I think this makes more sense, I will try it.

> * So unless I'm missing something, I would say there is no need to modify zerokit for the PoC you are developing in js-waku. One interesting abstraction could be to have a `generate_rln_proof_with_merkle_proof` that gets everything: merkle proof, message hash, epoch, secret, etc. So instead of requiring to i) generate witness and ii) generate rln proof with witness, you can do everything in just 1 function call. But again, a mere nice to have.
> 
> In case that helps, [here is](https://github.com/waku-org/go-zerokit-rln/blob/14960f3aff2a50088f780c233bda492ddbe26162/rln/rln_test.go#L372-L392) how I did it in go.

I agree, that would be ideal. If we move forward with this beyond PoC, I think it's worth considering.

adklempner:
Closing this as it was not necessary for the PoC

PR #234, opened 2024-03-29
add debug statements that appear in js console

adklempner:


tyshko-rostyslav:
Is this relevant, or should be closed?

PR #235, opened 2024-04-08
chore: remove code surface area for maintainability

rymnc:
These packages are no longer maintained, and increase the dependencies that we have. If these protocols do need further development, they can be re-added to the repo from this point in history.

Packages removed:
- private-settlement
- multiplier
- semaphore


rymnc:
the benchmarks comparing will fail, since it does a checkout of master, which is broken due to the semaphore dependency.

Issue #237, opened 2024-04-12
Release v0.5.0


rymnc:
With this issue we scope out from #91 the features and issues we want to address in the next zerokit release v0.5.0.

Reported items may change according to any eventual blocker found or re-prioritization of items.

## `maintenance`

- [x] Clean up the tests, refactor common tree init logic and insertion logic into macros or functions
- [x] Update documentation for rln-v2 to include new serde format 
- [x] add QoL traits to the Hasher associated type in MerkleTree trait: #238
- [x] refactor benchmarks, remove the following function https://github.com/vacp2p/zerokit/blob/0521c7349e7aa36e25b43e4663def226451a7147/rln/tests/poseidon_tree.rs#L13
- [x] Remove the `resources_folder` from the rln config

## `Usage of arkzkey vs zkey`

- [x] Include https://github.com/oskarth/mopro/tree/main/ark-zkey as part of the pre-processing of zkey to make the lib load faster

## `rln`

- [x] Remove `tree_height` 32 resources and associated circuit loading within https://github.com/vacp2p/zerokit/blob/d91a5b35688d7b033466dc879e494b07ecbbb6ee/rln/src/circuit.rs#L35-L37
- [x] Define the kats better in https://github.com/vacp2p/zerokit/blob/d91a5b35688d7b033466dc879e494b07ecbbb6ee/rln/tests/protocol.rs#L17-L69
- [x] Investigate tree reconciliation mechanism, where we can introspect and check which indices have 0 set in their leaves. If this is possible, patch rln-v1 branch with this feature too. This will help clients assess if their tree is structured appropriately
- [x] Expose a public function to fetch the root of a subtree at level `n`, i.e 
  ```rust 
  pub fn get_subtree_root<W: Write, R: Read>(&self, mut input_data: R, mut output_data: W) -> Result<()>
  ```

tyshko-rostyslav:
I'll take removing tree_height 32 resources

rymnc:
addressed in all the PRs mentioned above,
and released on crates.io, github too.

PR #239, opened 2024-04-22
Remove height 32 from RLN

tyshko-rostyslav:
Remove tree_height 32 resources and associated circuit loading within


rln/src/circuit.rs
```diff
@@ -33,8 +33,8 @@ const WASM_FILENAME: &str = "rln.wasm";
 // Note that the circuit and keys in TEST_RESOURCES_FOLDER are compiled for Merkle trees of height 20 & 32
 // Changing these parameters to other values than these defaults will cause zkSNARK proof verification to fail
 pub const TEST_PARAMETERS_INDEX: usize = 0;
-pub const TEST_TREE_HEIGHT: usize = [20, 32][TEST_PARAMETERS_INDEX];
-pub const TEST_RESOURCES_FOLDER: &str = ["tree_height_20", "tree_height_32"][TEST_PARAMETERS_INDEX];
+pub const TEST_TREE_HEIGHT: usize = [20][TEST_PARAMETERS_INDEX];
```
rymnc, line 36: you should also remove the tree_height_32 resources from the resources sub-directory.
please fix the comments which reference the tree height 32 too.
we can remove the TEST_TREE_HEIGHT, TEST_PARAMETERS_INDEX constants since they can have only one value.

rymnc (changes requested):
Thanks for this PR! Please look at the changes requested :)

rln/src/circuit.rs
tyshko-rostyslav: Removed resources.
Removed `TEST_PARAMETERS_INDEX `, but left `TEST_TREE_HEIGHT`, because it is used all over the place, and removing it would make it less descriptive.

tyshko-rostyslav:
Part of https://github.com/vacp2p/zerokit/issues/237

tyshko-rostyslav:
> Please also add some description to your PR, and link back to the issue in which this was assigned to you

Done

rymnc (approved):

PR #242, opened 2024-05-01
chore(rln): add ark-zkey support

seemenkina:
For now `makefile` doesn't work correct because of using whole `mopro` project as submodule. The problem is that `mopro` uses a patch for `wasmer` and `zerokit` doesn't and as a result there is a conflict in dependencies.

In general benchmark and tests are working if it runs by `cargo bench/tests`. 

So I opened an [issue](https://github.com/zkmopro/mopro/issues/131) to publish `ark-key` as a library and will wait for the result. In the meantime, I will try to pull up only part of the sub-module as a kludge solution and do some other tasks 

Part of #237


rln/src/public.rs
```diff
@@ -19,7 +19,8 @@ cfg_if! {
     if #[cfg(not(target_arch = "wasm32"))] {
         use std::default::Default;
         use std::sync::Mutex;
-        use crate::circuit::{circom_from_folder, vk_from_folder, circom_from_raw, zkey_from_folder, TEST_RESOURCES_FOLDER, TEST_TREE_HEIGHT};
+        use crate::circuit::{circom_from_folder, vk_from_folder, circom_from_raw, arkzkey_from_folder, TEST_RESOURCES_FOLDER, TEST_TREE_HEIGHT};
```
rymnc, line 22: perhaps there is a way we can maintain CI properly while merging this in, can we use arkzkey behind a feature flag? i.e behind the arkzkey feature flag we have the imports, loading logic etc, and in the default features we just continue to use zkey

rymnc (commented):
thanks for this PR! comments are inline

rln/src/public.rs
seemenkina: that sounds good, I'll add this one

rymnc:
perhaps we need to shallow clone the repo in the benchmarks github job/?

rln/src/circuit.rs
```diff
@@ -25,6 +23,19 @@ cfg_if! {
...
+        use ark_zkey::read_arkzkey_from_bytes;
+
+    } else {
+        use std::io::Cursor;
+        use ark_circom::read_zkey;
+    }
+}
+
+#[cfg(feature = "arkzkey")]
+const ARKZKEY_FILENAME: &str = "rln_final.arkzkey";
```
rymnc, line 37:
```suggestion
cfg_if! {
    if #[cfg(feature = "arkzkey")] {
        use ark_zkey::read_arkzkey_from_bytes;
        const ARKZKEY_FILENAME: &str = "rln_final.arkzkey";
    } else {
        use std::io::Cursor;
        use ark_circom::read_zkey;
    }
}
```
would this work?

seemenkina:
> perhaps we need to shallow clone the repo in the benchmarks github job/?

wouldn't that just make a dependency conflict? I mean between their patch wasmer and release wasmer?

rln/src/circuit.rs
```diff
@@ -101,6 +143,9 @@ pub fn vk_from_folder(resources_folder: &str) -> Result<VerifyingKey<Curve>> {
         ))?)?;
         Ok(verifying_key)
     } else if let Some(_zkey) = zkey {
+        #[cfg(feature = "arkzkey")]
+        let (proving_key, _matrices) = arkzkey_from_folder(resources_folder)?;
+        #[cfg(not(feature = "arkzkey"))]
         let (proving_key, _matrices) = zkey_from_folder(resources_folder)?;
```
rymnc, line 149: maybe we can abstract this loading into another function where we hide the
`#[cfg(feature = "arkzkey")]`

rln/src/circuit.rs
```diff
@@ -254,6 +299,9 @@ pub fn check_vk_from_zkey(
     resources_folder: &str,
     verifying_key: VerifyingKey<Curve>,
 ) -> Result<()> {
+    #[cfg(feature = "arkzkey")]
+    let (proving_key, _matrices) = arkzkey_from_folder(resources_folder)?;
+    #[cfg(not(feature = "arkzkey"))]
```
rymnc, line 304: same as above :)

rln/src/public.rs
```diff
@@ -158,6 +176,9 @@ impl RLN<'_> {
         #[cfg(not(target_arch = "wasm32"))]
         let witness_calculator = circom_from_raw(circom_vec)?;
 
+        #[cfg(feature = "arkzkey")]
+        let proving_key = arkzkey_from_raw(&zkey_vec)?;
+        #[cfg(not(feature = "arkzkey"))]
```
rymnc, line 181: same as above, this arkzkey/zkey abstraction should not leak into the public module

rymnc (commented):
Thanks for the changes, comments inline :)

rymnc:
> > perhaps we need to shallow clone the repo in the benchmarks github job/?
> 
> wouldn't that just make a dependency conflict? I mean between their patch wasmer and release wasmer?

yup, but we should try to find a way for the benchmarks to run in CI before we can merge this :)

rln/benches/pmtree_benchmark.rs
```diff
@@ -40,5 +50,21 @@ pub fn pmtree_benchmark(c: &mut Criterion) {
...
+    c.bench_function("ark_zkey::load", |b| {
+        b.iter(|| {
+            let _ = arkzkey_from_folder(TEST_RESOURCES_FOLDER);
+        })
+    });
+
+    #[cfg(not(feature = "arkzkey"))]
+    c.bench_function("zkey::load", |b| {
+        b.iter(|| {
+            let _ = rln::circuit::zkey_from_folder(TEST_RESOURCES_FOLDER);
```
seemenkina, line 64: maybe write in all possible calls like this? I mean if we have only one call in file we can avoid long definition of dependencies like in line 8-14 above

@rymnc  what do you think?

rln/src/circuit.rs
seemenkina: yep, I'll fix

rln/benches/pmtree_benchmark.rs
rymnc: hmm agree, also, sorry for the nit, maybe we can move this benchmark into a new file in the benches directory "circuit_loading_benchmarks.rs", since this is not directly related to the pmtree benchmarks

seemenkina: sure, np

rymnc:
> ## Benchmark for [88670db](https://github.com/vacp2p/zerokit/commit/88670dbb027b2a96543e74ee2e74a298c4a8eb5a)
> Click to view benchmark
> Test	Base	PR	%
> Pmtree::compute_root	0.0±0.00ns	0.0±0.00ns	NaN%
> Pmtree::get	324.0±1.78ns	**319.9±4.14ns**	**-1.27%**
> Pmtree::override_range	**230.0±3.32µs**	236.7±5.88µs	**+2.91%**
> Pmtree::set	55.8±0.42µs	**54.3±0.62µs**	**-2.69%**
> Pmtree:delete	55.7±0.43µs	**54.3±0.56µs**	**-2.51%**
> zkey::upload_from_folder	3.3±0.01s	N/A	N/A

this looks good -
can we add the reference to it in the rln crate's Cargo.toml?
```
[[bench]]
name = "circuit_loading_benchmark"
harness = false
```

PR #245, opened 2024-05-10
chore(rln) : Update documentation for rln-v2 to include new serde format

seemenkina:
Part of #237

### DIFF
--- a/rln/src/public.rs
+++ b/rln/src/public.rs
@@ -66,7 +66,7 @@ impl RLN<'_> {
     /// use std::io::Cursor;
     ///
     /// let tree_height = 20;
-    /// let resources = Cursor::new(json!({"resources_folder": "tree_config"});
+    /// let resources = Cursor::new(json!({ "resources_folder": TEST_RESOURCES_FOLDER }).to_string());;
     ///
     /// // We create a new RLN instance
     /// let mut rln = RLN::new(tree_height, resources);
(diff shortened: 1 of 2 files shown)


rln/src/public.rs
```diff
@@ -58,14 +58,15 @@ impl RLN<'_> {
...
-    ///
+    /// - `input_data`: include next parameters
+    ///     - `resources_folder`: a reader for the string path of the resource folder containing the ZK circuit (`rln.wasm`), the proving key (`rln_final.zkey`) or (`rln_final.arkzkey`) and the verification key (`verification_key.json`).
+    ///     - `tree_config`: a reader for a string containing a json with the merkle tree configuration
     /// Example:
     /// ```
     /// use std::io::Cursor;
     ///
     /// let tree_height = 20;
-    /// let resources = Cursor::new(json!({"resources_folder": "tree_height_20"});
```
rymnc, line 68: good catch, should we probably remove this as a config parameter? `resources_folder`

rymnc (approved):
LGTM, great stuff

can you add a sub-task to the v0.5.0 release issue which removes the `resources_folder` from the rln config?

rln/src/public.rs
seemenkina: hm, I guess if the data from our directory is always be used, then yes

PR #247, opened 2024-05-15
Expose a public function to fetch the root of a subtree at level n

seemenkina:
Part of #237


rymnc (approved):
LGTM, shall we wait for upstream new release on this one?

seemenkina:
Of course it doesn't work without this release. 

Or we need to make a fork on vac and depend on it while we wait for the release

Issue #248, opened 2024-05-16
Incorrect behaviour of trees in override_range function


seemenkina:
### 1. Full Merkle Tree and Optimal Merkle Tree 

This condition will not work correctly if the number of items to insert is less than the number of items to delete

https://github.com/vacp2p/zerokit/blob/4931b252378623bcc6f005164950710d152ee600/utils/src/merkle_tree/full_merkle_tree.rs#L181

At the same time the same behaviour is implemented correctly in `PmTree`:

https://github.com/vacp2p/zerokit/blob/4931b252378623bcc6f005164950710d152ee600/rln/src/pm_tree_adapter.rs#L287-L288

```
let leaves_new: Vec<TestFr> = (0..2).map(TestFr::from).collect();
tree.override_range(0, leaves_new.clone(), [0, 1, 2, 3]).unwrap();
```

### 2. PmTree 

Doesn't work correctly behaviour when we need to put leaves further away than deleted items because of this line: 

https://github.com/vacp2p/zerokit/blob/4931b252378623bcc6f005164950710d152ee600/rln/src/pm_tree_adapter.rs#L303-L304

and here should be `start` instead of `min_index`

For example: if we want to write data in indices [4..8] and delete data in indices [0..4], the data will be written from index 0 instead of 4. 

#### Example for testing: 

```
let leaves: Vec<Fr> = (0..4).map(|s| Fr::from(s as i32)).collect();
tree.override_range(4, leaves.clone(), [0, 1, 2, 3]).unwrap();
```

rymnc:
addressed in https://github.com/vacp2p/zerokit/pull/250

PR #249, opened 2024-05-16
Add function for checking indices of leaves which are set to zero

seemenkina:
Part of #237 

In general everything is ready, tests that helped to find problems from #248  are commented out for now.

### DIFF
--- a/rln/src/pm_tree_adapter.rs
+++ b/rln/src/pm_tree_adapter.rs
@@ -17,6 +17,9 @@ const METADATA_KEY: [u8; 8] = *b"metadata";
 
 pub struct PmTree {
     tree: pmtree::MerkleTree<SledDB, PoseidonHash>,
+    /// The indices of leaves which are set into zero upto next_index.
+    /// Set to 0 if the leaf is empty and set to 1 in otherwise.
+    cached_leaves_indices: Vec<u8>,
     // metadata that an application may use to store additional information
     metadata: Vec<u8>,
 }
@@ -144,6 +147,7 @@ impl ZerokitMerkleTree for PmTree {
 
         Ok(PmTree {
             tree,
+            cached_leaves_indices: vec![0; 1 << depth],
             metadata: Vec::new(),
         })
     }
@@ -156,7 +160,7 @@ impl ZerokitMerkleTree for PmTree {
         self.tree.capacity()
     }
 
-    fn leaves_set(&mut self) -> usize {
+    fn leaves_set(&self) -> usize {
         self.tree.leaves_set()
     }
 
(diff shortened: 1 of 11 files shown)


rln/src/pm_tree_adapter.rs
```diff
@@ -170,23 +174,41 @@ impl ZerokitMerkleTree for PmTree {
...
 
+    fn get_empty_leaves_indices(&self) -> Vec<usize> {
+        let next_idx = self.leaves_set();
+        self.cached_leaves_indices
+            .iter()
+            .take(next_idx)
+            .enumerate()
+            .filter(|&(_, &v)| v == 0u8)
+            .map(|(idx, _)| idx)
+            .collect()
```
rymnc, line 209: not sure if this is the best way to handle finding zero elements
perhaps we can employ a bloom filter and set values there?
quick napkin math suggests that we may use just ~3.5mb of memory vs 33mb

vec approach: 1048576 elements * 32 = 33mb at the cost of good performance
bloom filter: `m = ceil((n * log(p)) / log(1 / pow(2, log(2))))` where n = 1048576, p = 1000k = 3.59mb

rln/src/pm_tree_adapter.rs
```diff
@@ -16,6 +16,9 @@ const METADATA_KEY: [u8; 8] = *b"metadata";
 
 pub struct PmTree {
     tree: pmtree::MerkleTree<SledDB, PoseidonHash>,
+    /// The indices of leaves which are set into zero upto next_index.
+    /// Set to 0 if the leaf is empty and set to 1 in otherwise.
+    cached_leaves_indices: Vec<u8>,
```
rymnc, line 21: u8 probably not the best here, the highest index it can take is then 255, maybe should use usize instead or u32.

rymnc (commented):
Thanks for this, added comments in-line for alternative approach, and noticed the subtree test is included here, is that intended?

seemenkina:
> noticed the subtree test is included here, is that intended?

oh, It's just misprint in function name, thank you for noticed

rln/src/pm_tree_adapter.rs
seemenkina: it's not index itself - its 0 or 1 correspond the index to value 
 
 if we end up storing a value, then of course it will be usize

rln/src/pm_tree_adapter.rs
seemenkina: hm, I need to think about it 

If I think correctly, it will require rebuilding the filter every time we increase next_index? I'm not sure how much cost filter updating

seemenkina: I also wanna check false negative probability

rymnc (approved):
LGTM, thanks for this!

PR #251, opened 2024-05-20
Replace arkzkey git submodule to library

seemenkina:
mopro published a library release with arkzkey, so I replaced the submodule with that library 
Part of #237


rymnc (approved):
LGTM, thanks!

PR #255, opened 2024-05-21
feat: expose custom witness ffi

alrevuelta:
* Exposes `generate_rln_proof_with_witness` FFI, which allows to generate proofs with a custom witness.
* Similar as https://github.com/vacp2p/zerokit/pull/227/, which targeted RLN v1 but now targeting RLN v2.

### DIFF
--- a/rln/src/public.rs
+++ b/rln/src/public.rs
@@ -820,7 +820,7 @@ impl RLN<'_> {
     //
     // output_data is [ proof<128> | root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32>]
     // we skip it from documentation for now
-    #[doc(hidden)]
+    #[cfg(target_arch = "wasm32")]
     pub fn generate_rln_proof_with_witness<W: Write>(
         &mut self,
         calculated_witness: Vec<BigInt>,
--- a/rln/src/public.rs
+++ b/rln/src/public.rs
@@ -840,7 +840,7 @@ impl RLN<'_> {
 
     // Generate RLN Proof using a witness calculated from outside zerokit
     //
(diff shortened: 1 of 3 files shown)


rln/src/public.rs
```diff
@@ -820,7 +820,7 @@ impl RLN<'_> {
     //
     // output_data is [ proof<128> | root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32>]
     // we skip it from documentation for now
```
richard-ramos, line 822: This comment is no longer valid

richard-ramos (approved):

rln/src/public.rs
alrevuelta: thanks, addressed: c4ee1ff14c3b57879bccfc721d5bc4cfbbacc15b

rymnc (approved):
LGTM, ccing @seemenkina for another pair of eyes

rln/src/public.rs
```diff
@@ -839,6 +838,30 @@ impl RLN<'_> {
         Ok(())
     }
 
+    // Generate RLN Proof using a witness calculated from outside zerokit
+    //
+    // output_data is  [ proof<128> | root<32> | epoch<32> | share_x<32> | share_y<32> | nullifier<32> | rln_identifier<32> ]
```
seemenkina, line 843:
```suggestion
    // output_data is  [ proof<128> | root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32>]
```
it is little bit different format of output data now regarding the structure

seemenkina (approved):
LGTM

rln/src/public.rs
alrevuelta: indeed, good catch. fixed: 06393ae025da6c01257ac8c5aaf42d1ea2923583

PR #257, opened 2024-06-06
test(rln): use unchecked for Projective instantiation [dont merge]

rymnc:

PR #258, opened 2024-06-06
chore(rln): add verifying key deser benchmark

rymnc:
Benchmarking deserialization of verifying key. Need to merge this before https://github.com/vacp2p/zerokit/pull/257 to check if performance gains are being made.

### DIFF
--- a/rln/src/circuit.rs
+++ b/rln/src/circuit.rs
@@ -35,13 +35,13 @@ cfg_if! {
 }
 
 const ZKEY_FILENAME: &str = "tree_height_20/rln_final.zkey";
-const VK_FILENAME: &str = "tree_height_20/verification_key.json";
+pub const VK_FILENAME: &str = "tree_height_20/verification_key.json";
 const WASM_FILENAME: &str = "tree_height_20/rln.wasm";
 
 pub const TEST_TREE_HEIGHT: usize = 20;
 
 #[cfg(not(target_arch = "wasm32"))]
-static RESOURCES_DIR: Dir<'_> = include_dir!("$CARGO_MANIFEST_DIR/resources");
+pub static RESOURCES_DIR: Dir<'_> = include_dir!("$CARGO_MANIFEST_DIR/resources");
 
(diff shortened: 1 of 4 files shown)


rln/src/circuit.rs
```diff
@@ -248,28 +248,28 @@ fn json_to_g2(json: &Value, key: &str) -> Result<G2Affine> {
 }
 
 // Converts JSON to a VerifyingKey
-fn to_verifying_key(json: serde_json::Value) -> Result<VerifyingKey<Curve>> {
+pub fn to_verifying_key(json: &serde_json::Value) -> Result<VerifyingKey<Curve>> {
```
seemenkina, line 251: I suggest to use ark serilize: 

```
pub fn to_verifying_key(json: &serde_json::Value) -> Result<VerifyingKey<Curve>> {
    let data: Vec<u8> = serde::de::Deserialize::deserialize(json)?;
    let vk = VerifyingKey::<Curve>::deserialize_uncompressed(data.as_slice()).unwrap();
    Ok(vk)
}
```

Benchmark result: 
image

rln/src/circuit.rs
```diff
@@ -248,28 +248,28 @@ fn json_to_g2(json: &Value, key: &str) -> Result<G2Affine> {
...
 
 // Computes the verification key from its JSON serialization
 fn vk_from_json(vk: &str) -> Result<VerifyingKey<Curve>> {
     let json: Value = serde_json::from_str(vk)?;
-    to_verifying_key(json)
+    to_verifying_key(&json)
 }
 
 // Computes the verification key from a bytes vector containing its JSON serialization
 fn vk_from_vector(vk: &[u8]) -> Result<VerifyingKey<Curve>> {
```
seemenkina, line 268: Here we don't need to serialise into json at all if we will use ark deserialise

rln/src/circuit.rs
rymnc: right, I agree, however, this PR is just to add the benchmark *so* that it can be optimized. Will address it in a follow up PR.

seemenkina (approved):

PR #259, opened 2024-06-13
chore(rln): use ark serialized verification key for faster serde

rymnc:
Pre-processed the verification key in a similar manner to `rln_final.arkzkey`, and replaced the verification_key.json with it, and all references to it


rln/benches/circuit_deser_benchmark.rs
```diff
@@ -1,19 +1,17 @@
...
 // Here we benchmark how long the deserialization of the
 // verifying_key takes, only testing the json => verifying_key conversion,
 // and skipping conversion from bytes => string => serde_json::Value
 pub fn vk_deserialize_benchmark(c: &mut Criterion) {
     let vk = RESOURCES_DIR.get_file(Path::new(VK_FILENAME)).unwrap();
-    let vk = vk.contents_utf8().unwrap();
-    let json: Value = serde_json::from_str(vk).unwrap();
+    let vk = vk.contents();
 
     c.bench_function("circuit::to_verifying_key", |b| {
```
rymnc, line 12: leaving the bench_function name as-is to compare. will fix in following PR

rln/src/circuit.rs
```diff
@@ -161,115 +157,11 @@ pub fn circom_from_folder() -> Result<&'static Mutex<WitnessCalculator>> {
...
-
-// Computes the verification key from a bytes vector containing its JSON serialization
-fn vk_from_vector(vk: &[u8]) -> Result<VerifyingKey<Curve>> {
-    let json = String::from_utf8(vk.to_vec())?;
-    let json: Value = serde_json::from_str(&json)?;
-
-    to_verifying_key(&json)
+// Computes the verification key from a bytes vector containing pre-processed ark-serialized verification key
+// uncompressed, unchecked
+pub fn vk_from_slice(data: &[u8]) -> Result<VerifyingKey<Curve>> {
```
seemenkina, line 162: maybe we need other function name? Like `vk_from_ark_serilized` or `vk_from_ark_json` 

because I was expecting something along the lines of what we did when we pulled specific bytes from the slice

seemenkina (approved):
LGTM

rln/src/circuit.rs
rymnc: addressed in 400d9f1

PR #260, opened 2024-06-16
chore(rln): refactor resource initialization

rymnc:
Refactored the RLN module to use static resources instead of dynamically loaded resources, removing lifetime dependencies. Additionally, updated several RLN object methods and tests for compatibility with these changes.

### DIFF
--- a/rln/benches/circuit_loading_benchmark.rs
+++ b/rln/benches/circuit_loading_benchmark.rs
@@ -12,7 +12,7 @@ pub fn key_load_benchmark(c: &mut Criterion) {
 
 criterion_group! {
     name = benches;
-    config = Criterion::default().measurement_time(std::time::Duration::from_secs(250));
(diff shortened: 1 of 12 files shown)


rln/benches/circuit_loading_benchmark.rs
```diff
@@ -12,7 +12,7 @@ pub fn key_load_benchmark(c: &mut Criterion) {
 
 criterion_group! {
     name = benches;
-    config = Criterion::default().measurement_time(std::time::Duration::from_secs(250));
+    config = Criterion::default().without_plots().measurement_time(std::time::Duration::from_secs(10));
```
rymnc, line 15: need without plots, or the ci will fail with precision error on f64 values (too small time interval)

rymnc: ref: https://github.com/bheisler/criterion.rs/issues/388#issuecomment-1427653981

seemenkina (approved):
LGTM

PR #264, opened 2024-08-12
fix(rln-wasm): run tests again

rymnc:
Fixes regression introduced in #241

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,9 +71,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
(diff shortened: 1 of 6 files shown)


seemenkina:
in past [PR](https://github.com/vacp2p/zerokit/pull/253) the mismatch between serialisation and deserialisation (based on function names) has been corrected. Because the tests were not working it turned out that the functions were doing different things and I returned serialisation with big int to wasm, renamed it to a separate function so that there was no more such confusion

rymnc:
lgtm

Issue #263, opened 2024-07-29
Release v0.6.0


rymnc:
With this issue we scope out from #91 the features and issues we want to address in the next zerokit release v0.6.0.

Reported items may change according to any eventual blocker found or re-prioritization of items.

## `Usage of arkzkey`

- [x] compare circuit loading performance of a compressed arkzkey vs uncompressed arkzkey and report on the benchmark result, as well as the size of the file
- [x] instead of loading the zkey from file, investigate if the zkey can be defined in code, and if so, create a utility crate that accepts 2 inputs, one being the zkey from file, and the zkey defined in code, and compares the two, to return either a `true` or `false`. This will help in transparency as the result from the trusted setup is being defined differently.
- [ ] attempt a PoC with https://github.com/philsippl/circom-witness-rs instead of using the zkey. Benchmark and compare performance with the existing setup.
  - [ ] example: https://github.com/philsippl/semaphore-witness-example
  - [ ] rln-v2 circuits: https://github.com/Rate-Limiting-Nullifier/circom-rln

## `rln`
- [x] create a new feature `stateless` in the rln crate, which has the following requirements
  - [x] no merkle tree used
  - [x] exposes only 2 APIs, via FFI and the public module, one to [generate proofs](https://github.com/vacp2p/zerokit/blob/e6238fd7228dd645eb7d6f3bf2077a8623c38fcf/rln/src/public.rs#L845-L862) (already exists), and one to verify them. Utility functions may be exposed, like the poseidon hasher.
  - [x] Integrate `stateless` into rln-wasm, and subsequently, modify the rln-wasm api's to only include utilites, and to generate/verify proofs
 

## `rln-wasm`

- [ ] audit the current implementation to find performance pitfalls. 
- [ ] Bump dependencies of `wasmer` and benchmark against current implementation
- [ ] TBD: results of the `rln-wasm` audit

seemenkina:
Bottom line: using uncompressed arkzkey allows you to significantly reduce the program runtime when reading from a file. 
But there is a difficulty with the fact that the arkzkey library at this stage supports working only with compressed data, so I added an additional function to allow reading uncompressed keys 

See PR #269

<img width="728" alt="image" src="https://github.com/user-attachments/assets/83736a41-dc51-4766-a87b-352887eac396">
<img width="749" alt="image" src="https://github.com/user-attachments/assets/569e050a-c93e-4f9e-a4a3-9421a656a228">

seemenkina:
[...]

PR #265, opened 2024-08-15
Stateless Feature

seemenkina:
UPD: Ready for review, integration into rln-wasm will be in follow PR 

Based on issues for [release 0.6.0](https://github.com/vacp2p/zerokit/issues/263)

Main changes: 

- Now when the flag `stateless` is specified, the tree for rln will be unavailable
- Added new constructors + took into consideration whether wasm32 is enabled or not :  
https://github.com/vacp2p/zerokit/blob/46dac03701549f59d8abb37a9d7afed51e38dba5/rln/src/public.rs#L118-L119 and https://github.com/vacp2p/zerokit/blob/46dac03701549f59d8abb37a9d7afed51e38dba5/rln/src/public.rs#L213-L214 and https://github.com/vacp2p/zerokit/blob/46dac03701549f59d8abb37a9d7afed51e38dba5/rln/src/public.rs#L265-L266
- The function to generate the proof without using tree was already written so I used it without changing the name https://github.com/vacp2p/zerokit/blob/46dac03701549f59d8abb37a9d7afed51e38dba5/rln/src/public.rs#L935
- There is a function for verification without tree in rln too, also used it without changing the name: https://github.com/vacp2p/zerokit/blob/46dac03701549f59d8abb37a9d7afed51e38dba5/rln/src/public.rs#L1058

- All tree functions inside public and ffi are hidden under the condition that the `stateless` flag is not set 
- Split tests into submodules to allow testing of functionality with/without trees

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -8,7 +8,7 @@ jobs:
   linux:
     strategy:
       matrix:
-        feature: ["default", "arkzkey"]
+        feature: ["default", "arkzkey", "stateless"]
         target:
           - x86_64-unknown-linux-gnu
           - aarch64-unknown-linux-gnu
@@ -47,7 +47,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        feature: ["default", "arkzkey"]
+        feature: ["default", "arkzkey", "stateless"]
         target:
           - x86_64-apple-darwin
           - aarch64-apple-darwin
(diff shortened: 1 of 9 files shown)


rymnc:
nice
cc: @alrevuelta if you want to test with go-waku-light when this is ready for review

rymnc:
please can you also add this to the list of generated artifacts in the nightly release workflow?

seemenkina:
> please can you also add this to the list of generated artifacts in the nightly release workflow?

sure 
btw I also can fix test error

tyshko-rostyslav (approved):

PR #269, opened 2024-09-13
chore(rln): add uncompressed key and benches

seemenkina:
Release v0.6.0 (issue  #263)

Regarding the use of arkzkey the results are as follows: 
- using uncompressed version roughly reduces program runtime by a factor of (bench results: https://github.com/vacp2p/zerokit/issues/263#issuecomment-2348730752 )
- while defining the key in the code does not give any time gain, but at the same time it heavily loads RAM because the data size is too large to simply define it in files. 

As a result, leave reading from file, but use uncompressed arkzkey


rln/benches/circuit_loading_benchmark.rs
```diff
@@ -1,21 +1,24 @@
+use ark_circom::read_zkey;
 use criterion::{criterion_group, criterion_main, Criterion};
-use rln::circuit::{zkey_from_raw, ZKEY_BYTES};
+use std::io::Cursor;
 
-// Depending on the key type (enabled by the `--features arkzkey` flag)
-// the upload speed from the `rln_final.zkey` or `rln_final.arkzkey` file is calculated
-pub fn key_load_benchmark(c: &mut Criterion) {
-    let zkey = ZKEY_BYTES.to_vec();
+pub fn zkey_load_benchmark(c: &mut Criterion) {
```
tyshko-rostyslav, line 5: Any docs as before

tyshko-rostyslav (approved):

Issue #271, opened 2024-10-04
Draft: Release v0.6.1


seemenkina:
The scope of features and bugs that will be fixed in the minor v0.6.1 release

### Poseidon Hash
- [ ] Generate Poseidon round parameters and constants:
  - https://github.com/vacp2p/zerokit/issues/50 
  Only round parameters (easy) and secure MDS matrices verification algorithms (hard) missing
  - ensure that known answer tests match with the result of another implementation

## `Docs`
- [ ] https://github.com/vacp2p/zerokit/issues/256

AlekseiVambol:
Using the **arkzkey witness calculator** in zerokit implicitly involves running the corresponding **WASM** code. This circumstance suggests the possibility of **more performant witness computation**. Two alternative Circom witness calculators, which does not rely on WASM, were proposed recently:
[circom-witness-rs](https://github.com/philsippl/circom-witness-rs)
[circom-witnesscalc](https://github.com/iden3/circom-witnesscalc)

Thus, in order to improve the performance of the witness calculation process of zerokit in reasonable amount of time, these two alternatives have been studied in terms of **expected performance and convenience for integration** into zerokit. The study methods were static **code analysis** and review of **available documentation**.

This comparative study has yielded the following results:
1. **Both** calculators rely only on **Rust** code for the witness computation phase, which uses the execution graph. **No** evidence of **parallelization** has been found. See https://github.com/philsippl/circom-witness-rs/blob/fed7d591a4a6adfde371d3dac3fde23d755bcd39/src/graph.rs#L124 and https://github.com/iden3/circom-witnesscalc/blob/5cb365b6e4d9052ecc69d4567fcf5bc061c20e94/src/graph.rs#L368
2. Both calculators use HashMaps with String keys to describe input values for the witness computation process and operate on the Fp<MontBackend<FrConfig, 4>, 4> variables to compute and store the witness data. The later data type is provided by Arkworks.
3. [circom-witnesscalc](https://github.com/iden3/circom-witnesscalc) has been created by **iden3** (the creators of Circom) **on the basis of** [circom-witness-rs](https://github.com/philsippl/circom-witness-rs) "as a **drop-in replacement** for Circom's witness generator" and implements **more** execution graph **operations**. See the code references in the point 1.
[...]

PR #273, opened 2024-12-19
Replace the ark-zkey witness calculator with the one of iden3

AlekseiVambol:
**Motivation** 
See [this comment](https://github.com/vacp2p/zerokit/issues/271#issuecomment-2543981950).

**Methods**
1. Use [circom-witness-rs](https://github.com/philsippl/circom-witness-rs) to create the execution graph file for the [rln circuit](https://github.com/Rate-Limiting-Nullifier/circom-rln) used in zerokit. Add this file to the zerokit resources.

2. Remove the execution graph generator code from [circom-witness-rs](https://github.com/philsippl/circom-witness-rs) and simplify its architecture to facilitate integration of its witness calculating code into zerokit.

3. Rewrite the corresponding parts of the cut version of [circom-witness-rs](https://github.com/philsippl/circom-witness-rs) to make its input and output data format consistent with zerokit.

4. Integrate the modified cut version of [circom-witness-rs](https://github.com/philsippl/circom-witness-rs) into zerokit, remove the code expressing the involvement of the ark-zkey witness calculator and WASM binary file used by it, involve the integrated code into the witness calculation and remove the redundant dependencies.

5. Modify the code related to testing and measurements to facilitate studying the performance of the witness calculation for the release compilation mode.

**Results**
1. On the same computer the observed witness calculation times lied in **[1.72 ms, 3.49 ms]** for the proposed zerokit version and in **[11.42ms, 32.81 ms]** for the one using  the ark-zkey witness calculator. The precise comparative performance study is out of the scope of the current work.
2. Dependence on the **WASM binary** file has been **eliminated**. This facilitates using zerokit as a software component.


rln/src/iden3calc/storage.rs
```diff
@@ -0,0 +1,493 @@
...
+    if !magic.eq(WITNESSCALC_GRAPH_MAGIC) {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            "Invalid magic",
+        ));
+    }
+
+    let mut nodes = Vec::new();
+    let nodes_num = br.read_u64::<LittleEndian>()?;
+    for _ in 0..nodes_num {
```
tyshko-rostyslav, line 248: Probably a one-liner using collect would be more idiomatic

rln/src/iden3calc/storage.rs
```diff
@@ -0,0 +1,493 @@
...
+    Ok(())
+}
+
+fn read_message_length<R: Read>(rw: &mut WriteBackReader<R>) -> std::io::Result<usize> {
+    let mut buf = [0u8; MAX_VARINT_LENGTH];
+    let bytes_read = rw.read(&mut buf)?;
+    if bytes_read == 0 {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::UnexpectedEof,
+            "Unexpected EOF",
```
tyshko-rostyslav, line 198: Probably better to have these literals by name, reuse them and put them in a different file. Even better would be a more holistic approach to error handling, but that could be topic of another PR

rln/src/iden3calc/storage.rs
```diff
@@ -0,0 +1,493 @@
...
+fn read_message<R: Read, M: Message + std::default::Default>(
+    rw: &mut WriteBackReader<R>,
+) -> std::io::Result<M> {
+    let ln = read_message_length(rw)?;
+    let mut buf = vec![0u8; ln];
+    let bytes_read = rw.read(&mut buf)?;
+    if bytes_read != ln {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::UnexpectedEof,
+            "Unexpected EOF",
```
tyshko-rostyslav, line 222: Same as above

rln/src/public.rs
```diff
@@ -153,36 +139,33 @@ impl RLN {
...
     ///     let mut file = File::open(&fullpath).expect("no file found");
     ///     let metadata = std::fs::metadata(&fullpath).expect("unable to read metadata");
     ///     let mut buffer = vec![0; metadata.len() as usize];
     ///     file.read_exact(&mut buffer).expect("buffer overflow");
     ///     resources.push(buffer);
-    ///     let tree_config = "{}".to_string();
-    ///     let tree_config_input = &Buffer::from(tree_config.as_bytes());
     /// }
     ///
+    /// let tree_config = "{}".to_string();
```
tyshko-rostyslav, line 151: Are these lines needed?

rln/src/iden3calc/graph.rs
```diff
@@ -0,0 +1,957 @@
...
+/// Value numbering
+pub fn value_numbering(nodes: &mut Vec<Node>, outputs: &mut [usize]) {
+    assert_valid(nodes);
+
+    // Evaluate the graph in random field elements.
+    let values = random_eval(nodes);
+
+    // Find all nodes with the same value.
+    let mut value_map = HashMap::new();
+    for (i, &value) in values.iter().enumerate() {
```
tyshko-rostyslav, line 591: All of these for cycles below are not very idiomatic

rln/src/iden3calc.rs
```diff
@@ -0,0 +1,73 @@
...
+
+fn populate_inputs(
+    input_list: &HashMap<String, Vec<U256>>,
+    inputs_info: &InputSignalsInfo,
+    input_buffer: &mut [U256],
+) {
+    for (key, value) in input_list {
+        let (offset, len) = inputs_info[key];
+        if len != value.len() {
+            panic!("Invalid input length for {}", key);
```
tyshko-rostyslav, line 59: What is the justification for a panic?

rln/src/iden3calc/graph.rs
```diff
@@ -0,0 +1,957 @@
...
+                Ok(op.eval(self.to_const(NodeIdx(*a))?, self.to_const(NodeIdx(*b))?))
+            }
+            Node::TresOp(op, a, b, c) => Ok(op.eval(
+                self.to_const(NodeIdx(*a))?,
+                self.to_const(NodeIdx(*b))?,
+                self.to_const(NodeIdx(*c))?,
+            )),
+            Node::Input(_) => Err(NodeConstErr::InputSignal),
+            Node::MontConstant(_) => {
+                panic!("MontConstant should not be used here")
```
tyshko-rostyslav, line 318: What is the justification for a panic here?

rln/src/iden3calc/graph.rs
```diff
@@ -0,0 +1,957 @@
...
+            *a = renumber[*a].unwrap();
+            *b = renumber[*b].unwrap();
+            *c = renumber[*c].unwrap();
+        }
+    }
+    for output in outputs.iter_mut() {
+        *output = renumber[*output].unwrap();
+    }
+
+    eprintln!("Removed {removed} unused nodes");
```
tyshko-rostyslav, line 540: Maybe a good idea to think about a more holistic approach to logging

tyshko-rostyslav (commented):
Nothing critical, but I think it would be a good idea to fix a least some things I pointed out. If there is a rush another PR to fix it could be created and this one could be merged now.

rln/src/public.rs
AlekseiVambol: I think so. In the previous version these lines were within the loop body and therefore "tree_config" was not accessible by "new_with_params".

AlekseiVambol:
> Nothing critical, but I think it would be a good idea to fix a least some things I pointed out. If there is a rush another PR to fix it could be created and this one could be merged now.

Thank you for reviewing! Most fragments of the code you mentioned have been written by iden3 and can be found there: https://github.com/philsippl/circom-witness-rs . Thus, I think it is better to contact them regarding your remarks, and then the updated code from iden3 repository can be integrated. So I have responded to the comment regarding the part of code rewritten by me.

rln/README.md
```diff
@@ -12,13 +12,17 @@ git clone https://github.com/vacp2p/zerokit.git
...
-```bash
-cargo make build
-cargo make test
+``` bash
+ cargo make build
+ cargo make test
+```
+ ### Currently the tests are run as follows:
+ ``` bash
+cargo make test_default
```
seemenkina, line 25:
```
 cargo make build
```

Several builds are provided for tests based on the features: 

#### Default tests

 ``` bash
cargo make test_default
```

#### Tests with arkzkey feature
```
cargo make test_arkzkey
```

#### Tests with stateless feature
```
cargo make test_stateless

seemenkina: It's better not to use `Currently the tests are run as follows`. Let's make it like above

rln/src/iden3calc.rs
seemenkina: I agree, we need to get the results back, not panic

rln/src/iden3calc.rs
```diff
@@ -0,0 +1,73 @@
...
+    inputs: I,
+    graph_data: &[u8],
+) -> Vec<Fr> {
+    let inputs: HashMap<String, Vec<U256>> = inputs
+        .into_iter()
+        .map(|(key, value)| (key, value.iter().map(|v| U256::from(v)).collect()))
+        .collect();
+
+    let (nodes, signals, input_mapping): (Vec<Node>, Vec<usize>, InputSignalsInfo) =
+        deserialize_witnesscalc_graph(std::io::Cursor::new(graph_data)).unwrap();
```
seemenkina, line 27: Add error return, not unwrap

rln/src/iden3calc.rs
```diff
@@ -0,0 +1,73 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
```
seemenkina, line 3: What the reason oh having this one here? Why not fix names regarding linter?

rln/src/protocol.rs
```diff
@@ -544,13 +541,13 @@ pub fn generate_proof_with_witness(
     proving_key: &(ProvingKey<Curve>, ConstraintMatrices<Fr>),
 ) -> Result<ArkProof<Curve>, ProofError> {
     // If in debug mode, we measure and later print time take to compute witness
-    #[cfg(debug_assertions)]
+    #[cfg(test)]
```
seemenkina, line 544: Not sure that is good idea to replace this on test. They have different behaviour. Why are you change it?

rln/src/public.rs
seemenkina: btw, it's wrong config definition. It should be empty string 
```
let tree_config = "".to_string();
```

seemenkina (commented):
1. Agree about errors handling - no panics and unwrap, use `thiserror as in all code
2. Fix clippy error 
3. Return wasm file into resources - rln_wasm right now depend on this file especially in tests. And also check that you don't remove functionality around rln-wasm

seemenkina:
> Thank you for reviewing! Almost all the fragments of the code you mentioned have been written by iden3 and can be found there: https://github.com/philsippl/circom-witness-rs . Thus, I think it is better to contact them regarding your remarks, and then the updated code from iden3 repository can be integrated. So I have responded to the comment regarding the part of code rewritten by me.

As for me, all remarks are on point and it's worth fixing them because they affect the general style of the code. Just because the code is taken from another source doesn't mean it can't be changed, especially since it's taken without pulling dependencies (by the way, why not?).

rln/src/public.rs
AlekseiVambol: Are you sure? The previous version of the code contained "let tree_config = "{}".to_string();" within the loop body.

rln/src/iden3calc.rs
AlekseiVambol: Ok, I will remove this, since I do not know the motivation behind this.

rln/src/iden3calc.rs
AlekseiVambol: Agree.

rln/src/protocol.rs
AlekseiVambol: This is in accordance with "5. Modify the code related to testing and measurements to facilitate studying the performance of the witness calculation for the **release compilation mode**." Measurements in the debug mode in general may yield more pessimistic results compared to the real ones.

rln/src/iden3calc/storage.rs
AlekseiVambol: Agree.

rln/src/iden3calc/storage.rs
AlekseiVambol: No need in this, because the iden3 code contains only 2 such literals, which are used only 3 times, and the original zerokit code uses the same approach to error handling.
изображение

rln/src/iden3calc/graph.rs
AlekseiVambol: The cycle with HashMap does not need modifications, see the second part of https://stackoverflow.com/a/30441736 . The only loop in this function to be eliminated is the one at the line 596.

rln/src/iden3calc/graph.rs
AlekseiVambol: I guess there is no need in these logging parts, so I am going to remove them.

rln/src/iden3calc/graph.rs
AlekseiVambol: See https://github.com/vacp2p/zerokit/pull/273#discussion_r1895906160

tyshko-rostyslav:
> > Nothing critical, but I think it would be a good idea to fix a least some things I pointed out. If there is a rush another PR to fix it could be created and this one could be merged now.
> 
> Thank you for reviewing! Almost all the fragments of the code you mentioned have been written by iden3 and can be found there: https://github.com/philsippl/circom-witness-rs . Thus, I think it is better to contact them regarding your remarks, and then the updated code from iden3 repository can be integrated. So I have responded to the comment regarding the part of code rewritten by me.

If we are copying any code as is from other repositories, I would suppose it is mandatory to put doc that specifies it, that would include links to files we are copying. There were issues in such occasions

AlekseiVambol:
Thanks for participating in the first round of reviewing and welcome to the second!

rln/src/iden3calc/storage.rs
seemenkina: Make sense. 

Btw, do we know the estimate for the number of nodes? I mean that we know the size and it would be nice to allocate memory to the vector at once, instead of copying it every time we push it. 

Like this: 

`let mut nodes = Vec::with_capacity(nodes_num as usize);`

rln/src/iden3calc/storage.rs
seemenkina: I still think that processing is necessary because, if you pay attention, all external functions that are called by rln have this handling and this one don't leave utils lib in this form 

Moreover, as far as I can see you use unwrap in all calls, which is not very good either: https://github.com/vacp2p/zerokit/blob/d8232a5b6eb89aba836c83531aa8bba55fe7f770/rln/src/iden3calc.rs#L24

rln/src/iden3calc/storage.rs
AlekseiVambol: > Make sense.
> 
> Btw, do we know the estimate for the number of nodes? I mean that we know the size and it would be nice to allocate memory to the vector at once, instead of copying it every time we push it.
> 
> Like this:
> 
> `let mut nodes = Vec::with_capacity(nodes_num as usize);`

Agree. I will add this.

seemenkina (approved):

PR #276, opened 2025-01-29
chore: Update dependencies to latest versions

seemenkina:
UPD: As a result it was decided to deprecate the wasm part, besides here also the makefile is fixed and so far the work with arkzkey is completely inside the code without using the library (todo: consider the option of returning to mopro )

----

**Main problem**: zerokit fixes many dependencies and some versions are too old and new libraries don't support them and a conflict happened 

the main conflict is around `once_cell` and `num_traits`

`once_cell` - already fixed 

`num_traits` -  cannot be fixed yet because this strict dependency came from `ark-zkey`, and there it in turn came from a fork on an old version of arc-circom where version 2.15 is fixed 

Open issue in ark-zkey: https://github.com/zkmopro/ark-zkey/issues/1


vivianjeng:
please check this https://github.com/zkmopro/ark-zkey/issues/1#issuecomment-2623367539
and the new crate is published 
https://crates.io/crates/ark-zkey/0.1.2

AlekseiVambol:
When "make installdeps" and "make test" are executed, which is specified in the main README file, errors occur. No other suspicious things have been found.

seemenkina:
> When "make installdeps" and "make test" are executed, which is specified in the main README file, errors occur. No other suspicious things have been found.

could you provide log for `make installdeps`, because on my machine it works correctly 

for `make test`  it makes sense, I'll update makefile

AlekseiVambol:
> > When "make installdeps" and "make test" are executed, which is specified in the main README file, errors occur. No other suspicious things have been found.
> 
> could you provide log for `make installdeps`, because on my machine it works correctly
> 
> for `make test` it makes sense, I'll update makefile

The part containing warnings and errors:
изображение

seemenkina:
@AlekseiVambol I've fixed problem with `make test`

AlekseiVambol:
> @AlekseiVambol I've fixed problem with `make test`

Currently, the tests do not produce errors. No other problems have been found.

AlekseiVambol (approved):

PR #277, opened 2025-01-30
Update LICENSE-MIT

maximevtush:
Updated the copyright year to 2025 in the LICENSE file


seemenkina:
@maximevtush thank you for your input, but this edit does not make sense, because in the mit licence it is enough to specify the year of applied 

we look forward to your contribution to our other issues

PR #278, opened 2025-02-03
chore(build): improve WABT installation across platforms

seemenkina:
Modify Makefile to handle WABT installation consistently for macOS and Linux, ensuring proper setup and build of the WebAssembly Binary Toolkit (WABT).


seemenkina:
Move into https://github.com/vacp2p/zerokit/pull/276 because requires solve wasm problems

PR #280, opened 2025-02-13
fix(rln-cli): improve configuration handling, add relay and stateless example

vinhtc27:
Fix Configuration Handling:
+ Document the expected format of the configuration file.
+ Provide an example configuration file in rln-cli to guide users.
+ Ensure that the CLI provides useful error messages if the configuration is missing or incorrect.

Implement Relay and Stateless RLN Examples
+ Add Relay example to illustrate RLN’s practical use case.
+ Add Stateless RLN example, where an externally built Merkle tree provides paths and roots for RLN stateless feature.


rln-cli/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Configuration
+The configuration file is a JSON file that contains the configuration for the RLN instance.
+
+Example: [example.config.json](example.config.json)
+
+## Commands
+
+### Initialize a New RLN Instance
+```sh
+RLN_CONFIG_PATH=example.config.json cargo run new --config example.config.json --tree-height 20
```
seemenkina, line 13: why do we need set path twice?

rln-cli/src/main.rs
```diff
@@ -20,65 +25,72 @@ struct Cli {
...
+            resources_path,
             config,
-            tree_config_input,
         }) => {
             let mut resources: Vec<Vec<u8>> = Vec::new();
             #[cfg(feature = "arkzkey")]
             let filenames = ["rln_final.arkzkey", "verification_key.arkvkey"];
             #[cfg(not(feature = "arkzkey"))]
-            let filenames = ["rln_final.zkey", "verification_key.arkvkey"];
+            // let filenames = ["rln_final.zkey", "verification_key.arkvkey"];
```
seemenkina, line 51: Fix Cargo.toml instead of replace type of key

rln-cli/README.md
vinhtc27: Previously, State::load_state()? was required for all commands that needed RLN_CONFIG_PATH. 
I forgot to change that—I'll fix it

rln-cli/src/state.rs
```diff
@@ -12,12 +13,10 @@ pub(crate) struct State {
...
-            Some(RLN::new(tree_height, resources)?)
-        } else {
-            None
-        };
-        Ok(State { rln })
+        let resources = File::open(&config.path)?;
+
+        Ok(State {
+            rln: Some(RLN::new(config.tree_height, resources)?),
+        })
```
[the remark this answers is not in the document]
vinhtc27: I'll update it to work without a config file too

vinhtc27:
I force-pushed to replace two unsigned commits and will squash them into a single commit before the final review and merge

rln-cli/src/examples/stateless.rs
```diff
@@ -0,0 +1,214 @@
...
+    poseidon_tree::PoseidonTree,
+    protocol::{keygen, rln_witness_from_values, serialize_witness},
+    public::RLN,
+    utils::{bytes_le_to_fr, fr_to_bytes_le, normalize_usize},
+};
+use zerokit_utils::ZerokitMerkleTree;
+
+type ConfigOf<T> = <T as ZerokitMerkleTree>::Config;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
seemenkina, line 15: Maybe we can make similar example as for non-stateless RLN?

rln-cli/src/examples/relay.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        match Cli::try_parse_from(args) {
+            Ok(cli) => match cli.command {
+                Commands::List => {
+                    rln_system.list_users();
+                }
+                Commands::Register => {
+                    rln_system.register_user()?;
+                }
+                Commands::Send {
```
seemenkina, line 254: If I call this command twice with same message id it crashes

command: `send -u 0 -m 0 -s hi`
error: ```thread 'main' panicked at /Users/evb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ark-ff-0.4.2/src/fields/models/fp/mod.rs:907:42:
called `Option::unwrap()` on a `None` value```

rln-cli/src/main.rs
```diff
@@ -20,77 +29,93 @@ struct Cli {
...
-            state.rln = Some(RLN::new(*tree_height, resources)?);
+    match cli.command {
+        Some(Commands::New { tree_height }) => {
+            let config = Config::load_config()?;
+            state.rln = if let Some(InnerConfig { tree_height, .. }) = config.inner {
+                Some(RLN::new(tree_height, Cursor::new(config.as_bytes()))?)
+            } else {
+                Some(RLN::new(tree_height, Cursor::new(json!({}).to_string()))?)
+            };
             Ok(())
```
seemenkina, line 45: add here some simple comments that new instance created successfully

rln-cli/src/examples/relay.rs
seemenkina: hm, I'll check it

rln-cli/src/examples/relay.rs
```diff
@@ -0,0 +1,298 @@
...
+        Identity {
+            identity_secret_hash,
+            id_commitment,
+        }
+    }
+}
+
+struct RLNSystem {
+    rln: RLN,
+    // Mimic smart contract storage for used nullifiers for each epoch
```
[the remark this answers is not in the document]
vinhtc27: I was trying to illustrate how a smart contract might store used nullifiers and previous proof outputs in used_nullifiers field for the recover_id_secret function of RLN. 
I also used local_identities field to mimic how user secrets and public keys could be stored on each user’s device.
Any suggestions on a better way to represent this? Thanks!

rln-cli/src/examples/stateless.rs
vinhtc27: I think the relay.rs example already demonstrates the non-stateless RLN. 
I'll update the README for better clarification

rln-cli/src/examples/relay.rs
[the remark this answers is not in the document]
seemenkina: I'm not sure that it is a good idea to put it inside example staff. 

the reason is that I'm not sure it's a good example for production solutions in terms of efficiency and security. It is better to remove it from mentioning it in general, just note that for example we store data in these structures and one is needed for this and the other for that

rln-cli/src/examples/stateless.rs
seemenkina: no, I mean why stateless example looks different in terms of flow - it just more like test

rln-cli/src/examples/relay.rs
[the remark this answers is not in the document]
vinhtc27: Okay, I'll update the README and remove both comments to prevent any confusion

rln-cli/src/examples/stateless.rs
vinhtc27: Got it! I'll update the stateless example to be more like a CLI tool

rln-cli/src/examples/relay.rs
seemenkina: did you try to check through the RUST_BACKTRACE=1? 

So I found the problem - when you recover a secret, it actually assumes that we have two different signals for the same identifier and nullifiers. In the case where we send the same signal with the same parameters twice, there is no difference between the proofs and you divide by zero https://github.com/vacp2p/zerokit/blob/1930ca1610b85d59909fbbb6936ca042b6ebf1b8/rln/src/protocol.rs#L493

You need to add a check inside the ypur part that if the proof is the same, then you reject that message because it's duplicated and you can't even use it to recover the secret

rln-cli/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+The CLI supports optional features. To enable the **arkzkey** feature, run:
+
+```bash
+cargo run --features arkzkey -- <SUBCOMMAND> [OPTIONS]
+```
+
+For more details, refer to the [Zerokit RLN Module](../rln/README.md) documentation.
+
+## Relay Example
```
seemenkina, line 134: Let's put all this block before `## CLI Commands`

rln-cli/src/examples/relay.rs
```diff
@@ -0,0 +1,303 @@
...
+                    ))
+                }
+            }
+            Err(err) => Err(eyre!("Failed to recover identity secret: {err}")),
+        }
+    }
+}
+
+fn main() -> Result<()> {
+    let mut rln_system = RLNSystem::new()?;
```
seemenkina, line 234: When I run this command `cargo run --example relay` It takes a lot of time to load zkey (that's expected), but in cli it looks like it stuck. So I suggest to add here comment that you load data and create instance 

And same for stateless

seemenkina: could you also make the same in stateless file?

seemenkina (approved):
LGTM

PR #286, opened 2025-02-27
feat(rln): Add custom iden3 graph data support for RLN

seemenkina:
Brought back functionality that was removed during the migration to iden3 

- Modify circuit module to include graph data loading and calculation
- Update RLN struct to store graph data
- Adjust proof generation and FFI methods to use graph data
- Update benchmarks and tests to use sample size instead of measurement time
- Add new methods for graph data retrieval and handling


vinhtc27 (approved):
LGTM

PR #291, opened 2025-03-14
chore(rln): update dependencies and refactor code for compatibility

seemenkina:
While publishing the release on crate io it turned out that we can't use libraries without a version as it was for arc-circom. 

During the upgrade to the new version it was also discovered that it is possible to speed up input preparation for witness calculator by 4 times by switching from bigint to Fr:

image

it was also checked that it is also possible to use iden3 as a sub-module instead of copying code, but benchmarks showed that the new iden3 version with u256 calculations and subsequent conversion of the result to Fr is slower than the current implementation:

image

----

- Updated dependencies to their latest versions, including ark-ff, ark-bn254, ark-std, and others to 0.5.0.
- Refactored circuit and iden3calc modules to use Fr instead of BigInt for better type consistency.
- Improved utility functions for type conversions between Fr and U256.
- Adjusted Cargo.toml files for rln and utils to reflect new dependency versions and features.
- Enhanced documentation and comments for clarity on changes made.

This update ensures compatibility with the latest versions of the Ark framework and improves overall code quality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,7 +44,7 @@ dependencies = [
  "getrandom 0.2.15",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


vinhtc27 (approved):
LGTM

vinhtc27 (approved):

vinhtc27 (approved):

PR #292, opened 2025-03-17
feat(rln-wasm): porting the rln-wasm feature from v0.6.1 to v0.7.0

vinhtc27:
In zerokit rln-wasm module was built with wasmer, but the big problem with it is that it is difficult to support, very slow and problems with running calculations in parallel. The bottom line is to investigate other existing web support solutions and understand what can be used to deliver high performance on the web

### DIFF
--- a/rln-wasm/Cargo.toml
+++ b/rln-wasm/Cargo.toml
@@ -1,0 +1,36 @@
+[package]
+name = "rln-wasm"
+version = "0.1.0"
+edition = "2021"
+license = "MIT or Apache2"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+required-features = ["stateless"]
+
+[dependencies]
+rln = { path = "../rln", default-features = false }
+num-bigint = { version = "0.4.6", default-features = false, features = [
+    "rand",
+    "serde",
+] }
+web-sys = { version = "0.3.77", features = ["console"] }
+wasm-bindgen = "0.2.100"
+serde-wasm-bindgen = "0.6.5"
+js-sys = "0.3.77"
+serde_json = "1.0.140"
+
+# The `console_error_panic_hook` crate provides better debugging of panics by
+# logging them with `console.error`. This is great for development, but requires
+# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
+# code size when deploying.
+console_error_panic_hook = { version = "0.1.7", optional = true }
+zerokit_utils = { path = "../utils" }
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.13"
+wasm-bindgen-futures = "0.4.33"
+
+[features]
+default = ["console_error_panic_hook"]
+stateless = ["rln/stateless"]
--- a/rln-wasm/Makefile.toml
+++ b/rln-wasm/Makefile.toml
@@ -1,0 +1,31 @@
+[tasks.pack-build]
+command = "wasm-pack"
+args = ["build", "--release", "--target", "web", "--scope", "waku", "--features", "stateless"]
+
+[tasks.pack-rename]
+script = "sed -i.bak 's/rln-wasm/zerokit-rln-wasm/g' pkg/package.json && rm pkg/package.json.bak"
+
+[tasks.build]
+clear = true
+dependencies = ["pack-build", "pack-rename", "post-build"]
+
+[tasks.post-build]
+command = "wasm-strip"
+args = ["./pkg/rln_wasm_bg.wasm"]
+
+[tasks.test]
+command = "wasm-pack"
+args = ["test", "--release", "--node", "--target", "wasm32-unknown-unknown", "--features", "stateless"]
+dependencies = ["build"]
+
+[tasks.login]
+command = "wasm-pack"
+args = ["login"]
+
+[tasks.publish]
+command = "wasm-pack"
+args = ["publish", "--access", "public", "--target", "web"]
+
+[tasks.bench]
+command = "echo"
+args = ["'No benchmarks available for this project'"]
--- a/rln-wasm/README.md
+++ b/rln-wasm/README.md
@@ -1,0 +1,52 @@
+# RLN for WASM
+
+This library is used in [waku-org/js-rln](https://github.com/waku-org/js-rln/)
+
+## Building the library
+
+Install `wasm-pack`
+
+```bash
+curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+```
+
+Install `cargo-make`
+
+```bash
+cargo install cargo-make
+```
+
+OR
+
+```bash
+make installdeps
+```
+
+Compile zerokit for `wasm32-unknown-unknown`:
+
+```bash
+cd rln-wasm
+cargo make build
+```
+
+Compile a slimmer version of zerokit for `wasm32-unknown-unknown`:
+
+```bash
+cd rln-wasm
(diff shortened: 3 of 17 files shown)


rln/src/utils.rs
```diff
@@ -137,8 +137,10 @@ pub fn bytes_le_to_vec_usize(input: &[u8]) -> Result<Vec<usize>> {
 }
 
 #[inline(always)]
-pub fn normalize_usize(input: usize) -> [u8; 8] {
-    input.to_le_bytes()
+pub fn normalize_usize(input: usize) -> Vec<u8> {
```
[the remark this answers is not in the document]
vinhtc27: It was previously like this, and I changed it to [u8; 8] in the review-utils-func task. However, I realized that when enabling the wasm32-unknown-unknown target, utize is 4 bytes instead of 8, causing a compile error.

So, updating to the old way worked, and luckily, my change in v0.7.0 didn't cause any issues because there was no wasm32-unknown-unknown build.

rln/src/public.rs
```diff
@@ -223,6 +232,44 @@ impl RLN {
...
+    ///
+    /// let mut rln = RLN::new_with_params(
+    ///     resources[0].clone(),
+    ///     resources[1].clone(),
+    /// );
+    /// ```
+    #[cfg(all(target_arch = "wasm32", feature = "stateless"))]
+    pub fn new_with_params(zkey_vec: Vec<u8>) -> Result<RLN> {
+        let proving_key = zkey_from_raw(&zkey_vec)?;
+        let verification_key = vk_from_raw(&zkey_vec)?;
```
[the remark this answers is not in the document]
vinhtc27: Agree, I'll remove vk_from_raw and use .vk elsewhere as well.

rln/src/utils.rs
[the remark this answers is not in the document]
seemenkina: got it. But let's do like this to avoid heap allocation:

```
/// Normalizes a `usize` into an 8-byte array, ensuring consistency across architectures.
/// On 32-bit systems, the result is zero-padded to 8 bytes.
/// On 64-bit systems, it directly represents the `usize` value.
#[inline(always)]
pub fn normalize_usize(input: usize) -> [u8; 8] {
    let mut bytes = [0u8; 8]; 
    let input_bytes = input.to_le_bytes(); 
    bytes[..input_bytes.len()].copy_from_slice(&input_bytes);
    bytes
}
```

rln-wasm/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2.15", features = ["js"] }
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.50"
+wasm-bindgen-futures = "0.4.50"
+
+[features]
+default = ["console_error_panic_hook"]
+stateless = ["rln/stateless"]
```
seemenkina, line 38: put stateless as default because we don't have another option to use rln now

rln-wasm/Makefile.toml
```diff
@@ -0,0 +1,63 @@
...
+args = [
+    "test",
+    "--release",
+    "--node",
+    "--target",
+    "wasm32-unknown-unknown",
+    "--",
+    "--nocapture",
+]
+env = { "RUSTFLAGS" = "--cfg feature=\"stateless\"" }
```
seemenkina, line 37: check if it necessary when we move stateless feature as default

rln-wasm/README.md
```diff
@@ -0,0 +1,75 @@
...
+cargo make test
+```
+
+Or test with the **arkzkey** feature enabled
+
+```bash
+cargo make test_arkzkey
+```
+
+## Publishing an npm package
```
seemenkina, line 70: let's remove this part - it doesn't make sense for end-user

rln-wasm/README.md
```diff
@@ -0,0 +1,75 @@
...
+cargo make build
+```
+
+Or compile with the **arkzkey** feature enabled
+
+```bash
+cargo make build_arkzkey
+```
+
+Compile a slimmer version of zerokit for `wasm32-unknown-unknown`:
```
seemenkina, line 52: does it work and make sense?

rln-wasm/tests/rln-wasm.rs
```diff
@@ -0,0 +1,272 @@
...
+    const ZKEY_PATH: &str = "../rln/resources/tree_height_20/rln_final.zkey";
+
+    const CIRCOM_PATH: &str = "../rln/resources/tree_height_20/rln.wasm";
+
+    #[wasm_bindgen_test]
+    pub async fn rln_wasm_benchmark() {
+        let mut results = String::from("\nbenchmarks:\n");
+        let iterations = 10;
+
+        let zkey = read_file(&ZKEY_PATH).unwrap();
```
seemenkina, line 37: could you please go trough the rln-wasm test and replace unwrap() to expect(). Especially on RLN related function

rln/src/public.rs
```diff
@@ -81,8 +90,8 @@ impl RLN {
         )?;
 
         Ok(RLN {
-            proving_key: proving_key.to_owned(),
-            verification_key: verification_key.to_owned(),
+            proving_key,
+            verification_key,
             graph_data: graph_data.to_vec(),
```
seemenkina, line 95: let's also move `to_vec()` to graph data defintion

rln/src/public.rs
```diff
@@ -98,15 +107,15 @@ impl RLN {
...
+        let proving_key = zkey_from_folder().to_owned();
+        let verification_key = proving_key.0.vk.to_owned();
         let graph_data = graph_from_folder();
 
         Ok(RLN {
-            proving_key: proving_key.to_owned(),
-            verification_key: verification_key.to_owned(),
+            proving_key,
+            verification_key,
             graph_data: graph_data.to_vec(),
```
seemenkina, line 119: same

rln/src/public.rs
```diff
@@ -863,6 +904,28 @@ impl RLN {
         Ok(())
     }
 
+    // Generate RLN Proof using a witness calculated from outside zerokit
```
seemenkina, line 907: Make function description with  /// - we wanna this comments in docs. And add description about input value like in another functions here

rln-wasm/Cargo.toml
vinhtc27: I tried enabling stateless directly in rln-wasm without using a feature flag, but because we're in a workspace, cargo ends up activating stateless in other crates like rln-cli too — which breaks things, since rln-cli isn’t meant to work with stateless. 
So the safest way i found was to not make stateless a default feature, but instead gate it behind an explicit feature flag in rln-wasm, so we can control where it's enabled without affecting the rest of the workspace.

rln-wasm/README.md
vinhtc27: Did you mean removing the 'Publishing an npm package' part, or the 'arkzkey' feature part?

seemenkina: yes, `Publishing an npm package` part

rln-wasm/README.md
seemenkina: > Maybe we can just remove the wasm-strip command entirely, which would also eliminate the need for the wabt dependency in the installdeps command.


yeah, lets remove it

seemenkina (approved):
Good work, thank you!

PR #296, opened 2025-04-14
Enable parallel execution for rln-wasm module

vinhtc27:
## Changes

- Enabled parallelism in the browser for `rln-wasm` with the `multithread` feature flag.
- Added browser tests for both single-threaded and multi-threaded modes.
- Enabled browser tests in the CI workflow.
- Pending: resolving hanging issue with `wasm-bindgen-rayon` ([comment](https://github.com/RReverser/wasm-bindgen-rayon/issues/6#issuecomment-2814372940)).
- Forked [this commit](https://github.com/vacp2p/zerokit/tree/42887c80e65fee37310af91dd9d464b774d0cba6) into a separate [branch](https://github.com/vacp2p/zerokit/tree/benchmark-v0.8.0), which includes an HTML benchmark file and a test case for the multithreaded feature in `rln-wasm`. 
- The test case still has the known issue above, so it's temporarily disabled in this PR and will be addressed in the future.
- Improve the `make installdeps` which resolves the issue of NVM not enabling Node.js in the current terminal session.
- Reduce the build size of the `.wasm` blob using the `wasm-opt` tool from [Binaryen](https://github.com/WebAssembly/binaryen).
- Maybe we can close this draft [PR](https://github.com/vacp2p/zerokit/pull/226), which is already very outdated?


rln-wasm/Cargo.toml
```diff
@@ -37,3 +42,4 @@ wasm-bindgen-futures = "0.4.50"
 default = ["console_error_panic_hook"]
 stateless = ["rln/stateless"]
 arkzkey = ["rln/arkzkey"]
+rayon-parallelism = ["wasm-bindgen-rayon"]
```
seemenkina, line 45: let's choose a shorter name, like we used to - `parallel`

vinhtc27: I'll come up with a better feature name, but `parallel` doesn't work because it enables some internal dependency features that aren't supposed to be active.

seemenkina: hm, I got it, then it makes sense to get something else

rln-wasm/tests/browser.rs
```diff
@@ -0,0 +1,372 @@
...
+            "wasm_verify_with_roots: {}\n",
+            format_duration(wasm_verify_with_roots_result)
+        ));
+
+        // Log the results
+        console_log!("{results}");
+    }
+
+    #[wasm_bindgen_test]
+    pub async fn rln_wasm_test() {
```
seemenkina, line 259: Looks like test fully repeat benchmark flow. Do we actually need to separate it as we also check correctness in benchmarks and benchmarks run in the same flow with tests?

vinhtc27: I'll remove the redundant rln_wasm_test function

seemenkina (approved):
LGTM, thank you!

Issue #289, opened 2025-03-11
Release v0.8.0


seemenkina:
## `rln-wasm`
- [x] Bring back wasm support for zerokit
- [x] Investigate the possibility of using parallel computations 
- [ ] Update README 

## `rln`
- [x] Audit witness calculation part for RLN module
- [x] Audit atomic_operation function for tree structure 

## `docs`
- [ ] https://github.com/vacp2p/zerokit/issues/256
- [ ] https://github.com/vacp2p/zerokit/issues/303

vinhtc27:
# Zerokit `rln-wasm` module

## Background

Zerokit `rln-wasm` module was originally built with Wasmer, but it has significant drawbacks:

- Difficult to support due to compatibility issues.
- Very slow performance.
- Problems with parallel execution.

**Goal:** This release aims to explore alternative web-compatible solutions that deliver high performance.

## Current Status

- We have revived the old approach using the `ark-circom` crate (version `v0.5.0`) with the `wasm` feature enabled as a baseline for benchmarking and comparison. <https://github.com/vacp2p/zerokit/pull/292>
- This approach uses a `.wasm` file for witness calculation, with JS bindings that allow witness computation outside RLN.

## Challenges with the Current Approach

The current implementation introduces major compatibility issues:

- **Wasmer runtime**: Does not support 32-bit architectures. <https://github.com/wasmerio/wasmer/issues/4227#issuecomment-1738728892>
- **Dependency on `ark-circom`**: Vendor the latest `v0.5.0` release with `wasm` support requires a custom fork of `wasmer` due to dependency conflicts. <https://github.com/vinhtc27/wasmer>
- **Parallel execution not yet supported in the browser**: <https://github.com/vacp2p/zerokit/pull/109>

## Potential Solutions to Improve `rln-wasm`

As part of the transition away from `ark-circom` #282, we can consider the following improvements:

### 1. Witness Calculation

- We have migrated to `iden3` for witness calculation in the RLN module.
- Currently, `rln-wasm` uses JS bindings to `witness_calculator.js`, which is generated from `circom-rln` during circuit compilation.
- Performance could be improved if `circom-witnesscalc` gains native Wasm compatibility. <https://github.com/iden3/circom-witnesscalc/issues/10>

### 2. Key Parsing & Circom Reduction

[...]

PR #306, opened 2025-05-05
Enable parallel execution for Merkle Tree

vinhtc27:
### Changes

- Enhanced benchmark for testing parallelism.
- Enabled parallelism in both `OptimalMerkleTree` and `FullMerkleTree`.
- Added test cases to cover parallel execution and other operations.
- Benchmarked to determine the optimal `MIN_PARALLEL_NODES` threshold.
- Removed unused fields and updated method comments.
- Finalized benchmark report at [#289 in comment](https://github.com/vacp2p/zerokit/issues/289#issuecomment-2865107611).


vinhtc27:
Performed a force push to include the initial benchmark baseline (removed all previous benchmarks in this PR)

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -325,51 +326,39 @@ where
...
-            if parent_index >= parent_max_index {
-                // reset (aka decr depth & reset indexes)
-                parent_depth -= 1;
-                parent_index = parent_index_bak >> 1;
-                parent_index_bak = parent_index;
-                parent_max_index >>= 1;
-                current_depth -= 1;
-                current_index = current_index_bak >> 1;
-                current_index_bak = current_index;
+        let mut start = if index % 2 == 0 { index } else { index - 1 };
```
sydhds, line 330: Woud you mind keeping the old variable name (current_index)? "start" is very generic and we don't know what is inside.
Same for end, let's have an explicit name :)

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -325,51 +326,39 @@ where
...
+        let mut end = index + length;
+
+        while current_depth > 0 {
+            let parent_depth = current_depth - 1;
+
+            if end - start >= 64 {
```
sydhds, line 336: I'm wondering if this should be configurable or not. How did you set this value?

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -325,51 +326,39 @@ where
...
+                // Use Rayon parallelization when the level is wide enough
+                let updates: Vec<((usize, usize), H::Fr)> = (start..end)
+                    .step_by(2)
+                    .collect::<Vec<_>>()
+                    .into_par_iter()
```
sydhds, line 341: Any idea if the number of thread can be configured when using rayon? I'm wondering if this should be configurable by the user as well...

sydhds (changes requested):
Very nice, this is much more simple than the previous algorithm :)

utils/src/merkle_tree/optimal_merkle_tree.rs
vinhtc27: I’m currently benchmarking this to come up with a more general case and consider multiple factors to determine when to use sequential or parallel execution. I just put in a number here and made the benchmark more intensive to demonstrate the potential of this optimization.

utils/src/merkle_tree/optimal_merkle_tree.rs
vinhtc27: I’ll restore the previous convention for this function

utils/src/merkle_tree/optimal_merkle_tree.rs
vinhtc27: Rayon uses the optimal number of threads for the CPU by default. For example, if you have 8 cores, it will initialize 8 worker threads. The same number applies if you have 4 hyper-threaded cores. 
So, I think we should stick with Rayon’s default thread number.

seemenkina: I see it this way

1) if you have to enter additional config/environment variables for the configuration, then don't. We will make it too complicated to use for the end user

2) if it can only be done inside the code and dynamically defined by ourselves, then it makes sense

3) if it can be done with an existing environment variable, we can add this feature optionally + set default values for those who don't care.

utils/src/merkle_tree/optimal_merkle_tree.rs
seemenkina: Ideally, such a number should then be a constant and the docs should be left to make it clear what it is and where it comes from

sydhds (changes requested):

utils/src/merkle_tree/optimal_merkle_tree.rs
sydhds: So should we document the env var so user is aware there is a way to configure it?

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -276,100 +292,76 @@ where
...
+                    .map(|index| {
+                        // Hash two child nodes at positions (current_depth, index) and (current_depth, index + 1)
+                        let hash = self.hash_couple(current_depth, index);
+                        // Return the computed parent hash and its position at (parent_depth, index >> 1)
+                        ((parent_depth, index >> 1), hash)
+                    })
+                    .collect();
+
+                // Insert computed parent hashes into the tree
```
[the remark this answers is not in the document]
vinhtc27: Ill remove it

utils/tests/merkle_tree.rs
```diff
@@ -322,17 +406,64 @@ pub mod test {
...
 
-        // ensure that the leaves are set correctly
         for (i, &new_leaf) in new_leaves.iter().enumerate() {
-            assert_eq!(tree.get_leaf(i), new_leaf);
+            assert_eq!(tree_opt.get(i).unwrap(), new_leaf);
+        }
+    }
+
+    #[test]
+    fn test_override_range_parallel_triggered() {
```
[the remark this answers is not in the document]
vinhtc27: I'm just checking that everything still works correctly when run in parallel, the `leaf_count` just needs to be larger than PARALLEL_THRESHOLD.

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -262,7 +264,21 @@ where
     }
 
     fn compute_root(&mut self) -> Result<FrOf<Self::Hasher>> {
-        self.recalculate_from(0)?;
+        let mut i = 0;
```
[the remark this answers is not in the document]
vinhtc27: Yes, thanks for pointing that out, silly me!

utils/benches/merkle_tree_benchmark.rs
```diff
@@ -47,36 +46,57 @@ impl FromStr for TestFr {
...
+    static ref INDICES: Vec<usize> = (0..(1 << 20)).collect();
 }
 
 pub fn optimal_merkle_tree_benchmark(c: &mut Criterion) {
     let mut tree =
-        OptimalMerkleTree::<Keccak256>::new(2, TestFr([0; 32]), OptimalMerkleConfig::default())
+        OptimalMerkleTree::<Keccak256>::new(20, TestFr([0; 32]), OptimalMerkleConfig::default())
             .unwrap();
 
+    for i in 0..8192 {
```
seemenkina, line 66: Make constant variable for this number as it using not one during benchmarks

utils/benches/merkle_tree_benchmark.rs
```diff
@@ -47,36 +46,57 @@ impl FromStr for TestFr {
...
 lazy_static! {
-    static ref LEAVES: [TestFr; 4] = [
-        hex!("0000000000000000000000000000000000000000000000000000000000000001"),
-        hex!("0000000000000000000000000000000000000000000000000000000000000002"),
-        hex!("0000000000000000000000000000000000000000000000000000000000000003"),
-        hex!("0000000000000000000000000000000000000000000000000000000000000004"),
-    ]
-    .map(TestFr);
+    static ref LEAVES: Vec<TestFr> = {
+        let mut leaves = Vec::with_capacity(1 << 20);
```
seemenkina, line 50: I suggest to rename `TEST_TREE_HEIGHT` to `TEST_TREE_DEPTH` as it makes much more sense, @sydhds what do u think?

and anyway let's use constant  for number 20

utils/benches/merkle_tree_benchmark.rs
```diff
@@ -87,15 +107,21 @@ pub fn optimal_merkle_tree_benchmark(c: &mut Criterion) {
...
         })
     });
 
-    // check intermediate node getter which required additional computation of sub root index
     c.bench_function("OptimalMerkleTree::get_subtree_root", |b| {
+        let mut level = 1;
+        let mut index = 0;
         b.iter(|| {
-            tree.get_subtree_root(1, 0).unwrap();
+            tree.get_subtree_root(level % 19, index % (1 << (19 - (level % 19))))
```
seemenkina, line 121: why 19 here?

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -145,6 +144,7 @@ where
         Ok(self.nodes[self.capacity() + leaf - 1])
     }
 
+    // Returns the root of the subtree at level n and index
```
seemenkina, line 147: Maybe we can make this comments as docs (using ///)

utils/src/merkle_tree/optimal_merkle_tree.rs
seemenkina: Did I understand correctly that you are talking about an external env var?

I think so, at least a link to the docs from the lib itself

vinhtc27: Since `zerokit` is a library, i think we should not configure the global Rayon thread pool, by calling `ThreadPoolBuilder` or setting the `RAYON_NUM_THREADS` env variable. This configuration should be left to the root application, which has full control over global settings.

We can mention in our documentation that `zerokit` uses Rayon for parallelism, and users are responsible for any performance tuning or thread configuration if needed.

**Reference:** [Rayon FAQ – How many threads will Rayon spawn?](https://github.com/rayon-rs/rayon/blob/main/FAQ.md#how-many-threads-will-rayon-spawn)

utils/src/merkle_tree/optimal_merkle_tree.rs
```diff
@@ -276,100 +278,75 @@ where
...
+            // Use parallel processing when the number of pairs exceeds the threshold
+            if current_index_max - current_index >= MIN_PARALLEL_NODES {
+                let updates: Vec<((usize, usize), H::Fr)> = (current_index..current_index_max)
+                    .step_by(2)
+                    .collect::<Vec<_>>()
+                    .into_par_iter()
+                    .map(|index| {
+                        // Hash two child nodes at positions (current_depth, index) and (current_depth, index + 1)
+                        let hash = self.hash_couple(current_depth, index);
+                        // Return the computed parent hash and its position at (parent_depth, index >> 1)
```
seemenkina, line 330:
```suggestion
                        // Return the computed parent hash and its position
```
I think it's clear enough without repeating return value

utils/tests/merkle_tree.rs
```diff
@@ -83,21 +83,93 @@ pub mod test {
...
         }
     }
 
+    #[test]
+    fn test_compute_root() {
+        let depth = 4;
+        let leaves: Vec<TestFr> = (0..(1 << depth) as u32).map(TestFr::from).collect();
+
+        let mut tree_full = default_full_merkle_tree(depth);
+        tree_full.set_range(0, leaves.iter().cloned()).unwrap();
```
seemenkina, line 107: let's use `expect()` with comment instead of `unwrap()`

utils/src/merkle_tree/optimal_merkle_tree.rs
seemenkina: Agree

utils/src/merkle_tree/optimal_merkle_tree.rs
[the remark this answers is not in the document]
vinhtc27: Hey @sydhds, I just realized we can return self.root() in this method, just like in FullMerkleTree, since every modification in OptimalMerkleTree already calls update_hashes(). So I don't think we need to recompute the root here.
Maybe I'm missing something — is it an intentional design choice to avoid storing all nodes? But from what I understand, once something is inserted, the nodes are all stored in the structure anyway.

utils/benches/merkle_tree_benchmark.rs
vinhtc27: I wanted to benchmark all subtrees except the full tree, but the entire tree is a valid subtree too. I'll change it to 20 to cover all levels. Thanks!

utils/benches/merkle_tree_benchmark.rs
vinhtc27: The `TEST_TREE_HEIGHT` constant is from the `rln` module, right?

sydhds (approved):

utils/benches/merkle_tree_benchmark.rs
seemenkina: in general if you look at the implementation, the height of the tree in the tests you will have 21, 20 is the depth level (leaves level). It's not a bug, we just chose a bad name for the variable, we have already discussed this with @sydhds and it seems to me that maybe we should rename the constant to match the implementation.

i.e. rename `TEST_TREE_HEIGHT` to `TEST_TREE_DEPTH` in rln module

utils/tests/merkle_tree.rs
seemenkina: I mean entire file. It's better in case to catch any problem if we will update related code in future

vinhtc27: Got it

utils/benches/merkle_tree_benchmark.rs
vinhtc27: Can we leave this refactoring for another PR? It’s somewhat out of scope for this one. I suggest we create a separate task focused solely on:
  - Refactoring and renaming
  - Documentation
  - Reviewing the current feature flags related to MerkleTree usage in the rln module.
 
The current feature flags in rln are a bit confusing regarding which MerkleTree implementation is being used, and whether it is persisted to a file or not:

Current feature flags in `rln` is abit confusing about which MerkleTree is using, and does it persistant to file or not:
```
[features]
default = ["pmtree-ft"]
fullmerkletree = ["default"]
stateless = []
arkzkey = []

# Note: the pmtree feature is still experimental
pmtree-ft = ["utils/pmtree-ft"]
```

For example:
- rln-cli is currently using PmTree
- rln-wasm is using OptimalMerkleTree
I just realized these confusions may lead to inconsistencies when I started working on the MerkleTree task.

seemenkina: Sounds good to me, let's do like this

utils/src/merkle_tree/optimal_merkle_tree.rs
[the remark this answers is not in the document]
vinhtc27: Hey @seemenkina, me and @sydhds  were discussing the compute_root function in the ZerokitMerkleTree trait and we both agree it might not be necessary.

It doesn’t seem to be used outside of benchmarks, and the current implementations either just return self.root() (like in FullMerkleTree and PmTree) or perform redundant work (as in the earlier version of OptimalMerkleTree).

Do you think we could remove it to simplify the ZerokitMerkleTree trait?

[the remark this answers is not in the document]
seemenkina: Hm, looks like we can do it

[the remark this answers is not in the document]
vinhtc27: I removed the compute_root method from all MerkleTree types and updated the related tests accordingly.

seemenkina (approved):
LGTM

PR #318, opened 2025-06-12
Waku be funcs

sydhds:


rln/src/public.rs
```diff
@@ -1139,6 +1139,15 @@ impl RLN {
         Ok(())
     }
 
+    /// Same as key_gen but serialized in BE format
```
seemenkina, line 1142: btw, I think it's better to add the full documentation when we merge the PR. There's no need for it right now while it's still a WIP, but let's not forget about it

PR #319, opened 2025-06-12
Add IdSecret to wrap identity_secret_hash + Zeroize support

sydhds:


sydhds:
Close in favor of #320

PR #320, opened 2025-06-17
Initial impl for IdSecret

sydhds:


rln-cli/src/examples/relay.rs
```diff
@@ -221,20 +222,20 @@ impl RLNSystem {
...
                 {
-                    let real_identity_secret_hash = identity.identity_secret_hash;
+                    let real_identity_secret_hash = identity.identity_secret_hash.clone();
                     if leaked_identity_secret_hash != real_identity_secret_hash {
-                        Err(eyre!("identity secret hash mismatch {leaked_identity_secret_hash} != {real_identity_secret_hash}"))
+                        Err(eyre!("identity secret hash mismatch: leaked_identity_secret_hash != real_identity_secret_hash"))
                     } else {
-                        println!("DUPLICATE message ID detected! Reveal identity secret hash: {leaked_identity_secret_hash}");
+                        println!(
+                            "DUPLICATE message ID detected! Reveal identity secret hash: ********"
```
seemenkina: I think it makes sense to print the secret in the example, because we mostly use it for display, and if we can't show the secret on the screen, it would be weird.

rln/src/protocol.rs
```diff
@@ -618,7 +633,8 @@ pub fn inputs_for_witness_calculation(
         .for_each(|v| identity_path_index.push(Fr::from(*v)));
 
     Ok([
-        ("identitySecret", vec![rln_witness.identity_secret]),
+        // FIXME
```
seemenkina: what's wrong here?

rln/src/utils.rs
```diff
@@ -150,3 +155,66 @@ pub fn normalize_usize(input: usize) -> [u8; 8] {
 pub fn generate_input_buffer() -> Cursor<String> {
     Cursor::new(json!({}).to_string())
 }
+
+#[derive(
+    Debug,
+    Zeroize,
+    ZeroizeOnDrop,
+    // From,
```
seemenkina: lets's remove unused derives

rln/src/utils.rs
```diff
@@ -150,3 +155,66 @@ pub fn normalize_usize(input: usize) -> [u8; 8] {
...
+    }
+
+    pub(crate) fn to_bytes_le(&self) -> Zeroizing<Vec<u8>> {
+        let input_biguint: BigUint = self.0.into();
+        let mut res = input_biguint.to_bytes_le();
+        res.resize(fr_byte_size(), 0);
+        Zeroizing::new(res)
+    }
+
+    /*
```
seemenkina: same here

rln-cli/src/examples/relay.rs
vinhtc27: @sydhds 
I think it's fine to keep the code as-is and use the Display macro for IdSecret, since the secret key is just a random value in this example.

Zeroize only clears memory after it's no longer in use—it doesn't protect data while it's active. So keeping the example simple is better, and printing the secret isn't a problem here.

rln-cli/src/examples/stateless.rs
```diff
@@ -93,7 +94,7 @@ impl RLNSystem {
         println!("Registered users:");
         for (index, identity) in &self.local_identities {
             println!("User Index: {index}");
-            println!("+ Identity Secret Hash: {}", identity.identity_secret_hash);
+            println!("+ Identity Secret Hash: ********");
```
vinhtc27: Same goes for this example

rln/src/public.rs
vinhtc27, line 1: There are a few Clippy warnings in this file when using the stateless feature, such as with `cargo make test_stateless` on `rln` module or in the `rln-wasm` module.

rln/Cargo.toml
```diff
@@ -53,6 +53,8 @@ rand = "0.8.5"
 rand_chacha = "0.3.1"
 ruint = { version = "1.15.0", features = ["rand", "serde", "ark-ff-04"] }
 tiny-keccak = { version = "2.0.2", features = ["keccak"] }
+zeroize = "1.8"
+derive_more = { version = "2.0.1", features = ["from", "into", "display"] }
```
vinhtc27: Unuse deps here. We can implement `Display` for any type directly, instead of using a macro from another library, for a lighter binary.

rln/src/protocol.rs
sydhds: Pb here is that we are creating a copy of IdSecret (because *rln_witness.identity_secret create a copy of the underlying Fr) so it "leaks" the secret key here. Problem here is the return type of the function [(_, Vec<Fr>)] as we ideally should return the IdSecret here. Any idea of how to fix this?

rln/src/utils.rs
sydhds: Removed

rln/src/utils.rs
sydhds: Removed

rln/Cargo.toml
sydhds: Implementing Display directly will not make a binary lighter... derive_more is very handy for implementing Display (and many more traits) directly. Anyway I've removed it as it is not used

rln/src/protocol.rs
seemenkina: I'm not sure we can change anything.

the function is used just before the start of the witness calculation and there we need a field element already, not another type

I guess it makes sense to leave a comment here?

rln-cli/src/examples/relay.rs
sydhds: Re added

rln/src/protocol.rs
sydhds: I've added an enum and added some zeroize in iden3calc code. Let me know what you think about this :)

rln-cli/src/examples/stateless.rs
sydhds: Re added

rln/src/public.rs
sydhds: Should be fixed

seemenkina (approved):
LGTM, thank you!

rln-cli/src/examples/relay.rs
```diff
@@ -118,7 +119,7 @@ impl RLNSystem {
         match self.rln.set_next_leaf(&mut buffer) {
             Ok(_) => {
                 println!("Registered User Index: {index}");
-                println!("+ Identity secret hash: {}", identity.identity_secret_hash);
+                println!("+ Identity secret hash: ********");
```
vinhtc27: Just add this: print the identity_secret_hash for consistency like above, and I think it's good to merge!

rln-cli/src/examples/stateless.rs
```diff
@@ -107,7 +108,7 @@ impl RLNSystem {
         self.tree.update_next(rate_commitment)?;
 
         println!("Registered User Index: {index}");
-        println!("+ Identity secret hash: {}", identity.identity_secret_hash);
+        println!("+ Identity secret hash: ********");
```
vinhtc27: And here too

rln-cli/src/examples/stateless.rs
vinhtc27: And re added here too

sydhds: Sorry should be good now

rln-cli/src/examples/relay.rs
sydhds: Should be good now

rln-cli/src/examples/stateless.rs
sydhds: Fixed

sydhds:
> @sydhds There are also some Clippy warning introduced by the latest Rust 1.88 version at:
> 
> * `rln-cli/src/main.rs`
> * `rln/src/circuit/iden3calc.rs`
> 
> Since they're minor, i think we could address them in this PR as well.

The clippy warnings are about inlining the format args but not sure if we should follow this or not... Wdyt @seemenkina ?

vinhtc27 (approved):

seemenkina:
@sydhds just understand that current CI on previous rust version

so let's not change cargo clippy error on this PR

vinhtc27 (approved):

seemenkina (approved):

PR #325, opened 2025-07-09
feat: restore parallel flag, improve CI, resolve clippy warnings, bump deps

vinhtc27:
- Addressed the current default enablement of the parallel feature within the rln module.
- Updated all feature flags and command in rln-wasm from multithread to parallel.
- Resolve macro expansion problem inside `qap.rs` file by some Rayon’s trait.
- Tried to optimize workflow runtime with newer version of action packages.
- Resolved all new clippy warnings from Rust 1.88.
- Bump minor dependencies version.


.github/workflows/ci.yml
```diff
@@ -124,38 +112,35 @@ jobs:
           fi
         working-directory: ${{ matrix.crate }}
 
-  # rln-wasm-multihread-test:
+  # rln-wasm-parallel-test:
```
seemenkina, line 115: just in case, why this job still commented?

.github/workflows/ci.yml
```diff
@@ -169,32 +154,29 @@ jobs:
       matrix:
         # we run lint tests only on ubuntu
         platform: [ubuntu-latest]
-        crate: [rln, rln-wasm, utils]
+        crate: [rln, rln-cli, rln-wasm, utils]
```
seemenkina: do we really need to run tests for rln-cli on CI?

rln/Cargo.toml
```diff
@@ -67,7 +55,16 @@ document-features = { version = "0.2.11", optional = true }
 criterion = { version = "0.6.0", features = ["html_reports"] }
 
 [features]
-default = ["pmtree-ft"]
+default = ["parallel", "pmtree-ft"]
```
seemenkina, line 58: so that means we always run rln in parallel but have two options for rln-wasm and utils? 

in such case, how we can use rln in parallel in non parallel rln-wasm?

rln/tests/poseidon_tree.rs
```diff
@@ -15,17 +15,21 @@ mod test {
     // The test is checked correctness for `FullMerkleTree` and `OptimalMerkleTree` with Poseidon hash
```
seemenkina:
```suggestion
    // The test checked correctness for `FullMerkleTree` and `OptimalMerkleTree` with Poseidon hash
```

.github/workflows/ci.yml
vinhtc27: I’ll uncomment this when I can resolve the unit‑test problem related to the `wasm-bindgen-rayon` crate in the [rln-wasm-maintaining](https://github.com/vacp2p/roadmap/blob/acz-2025q3/content/acz/ift/2025q3-zerokit.md#description-5) task.

.github/workflows/ci.yml
vinhtc27: Yeah, it doesn't need the rln-cli

rln/Cargo.toml
vinhtc27: I updated how rln is used in rln-wasm:

```toml
rln = { path = "../rln", version = "0.8.0", default-features = false, features = ["stateless"] }
```

By default, rln-wasm uses the single-threaded version. 
I added a separate feature flag to optionally enable parallel when needed.

```
[features]
default = ["console_error_panic_hook"]
arkzkey = ["rln/arkzkey"]
parallel = ["rln/parallel", "wasm-bindgen-rayon"]
```

PR #329, opened 2025-07-25
feat: resolve overlap between stateless and merkletree feature flags

vinhtc27:
- Resolved overlap between stateless and merkletree feature flags.
- Updated every testcase related to stateless feature.
- Added compile-time feature check to avoid feature overlap.
- Added --no-default-features for all builds in nightly-release.yml [(tested)](https://github.com/vacp2p/zerokit/actions/runs/16525062203).

### DIFF
--- a/rln/src/lib.rs
+++ b/rln/src/lib.rs
@@ -13,15 +13,13 @@ pub mod public;
 pub mod public_api_tests;
 pub mod utils;
 
-// Feature validation for incompatible combinations
-#[cfg(all(feature = "fullmerkletree", feature = "optimalmerkletree"))]
-compile_error!("Cannot enable both fullmerkletree and optimalmerkletree");
-
-#[cfg(all(feature = "fullmerkletree", feature = "pmtree-ft"))]
-compile_error!("Cannot enable both fullmerkletree and pmtree-ft");
-
-#[cfg(all(feature = "optimalmerkletree", feature = "pmtree-ft"))]
-compile_error!("Cannot enable both optimalmerkletree and pmtree-ft");
+// Ensure that only one Merkle tree feature is enabled at a time
+#[cfg(any(
+    all(feature = "fullmerkletree", feature = "optimalmerkletree"),
+    all(feature = "fullmerkletree", feature = "pmtree-ft"),
+    all(feature = "optimalmerkletree", feature = "pmtree-ft"),
+))]
+compile_error!("Only one of `fullmerkletree`, `optimalmerkletree`, or `pmtree-ft` can be enabled at a time.");
 
 #[cfg(all(
     feature = "stateless",
--- a/rln/src/lib.rs
+++ b/rln/src/lib.rs
@@ -29,4 +29,4 @@ compile_error!("Only one of `fullmerkletree`, `optimalmerkletree`, or `pmtree-ft
         feature = "pmtree-ft"
     )
 ))]
-compile_error!("Cannot enable any merkletree features with stateless");
+compile_error!("Cannot enable any Merkle tree features with stateless");
--- a/rln/src/lib.rs
+++ b/rln/src/lib.rs
@@ -5,9 +5,30 @@ pub mod ffi;
 pub mod hashers;
 #[cfg(feature = "pmtree-ft")]
 pub mod pm_tree_adapter;
+#[cfg(not(feature = "stateless"))]
 pub mod poseidon_tree;
 pub mod protocol;
 pub mod public;
 #[cfg(test)]
 pub mod public_api_tests;
 pub mod utils;
+
+// Feature validation for incompatible combinations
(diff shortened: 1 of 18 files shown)


rln/src/lib.rs
```diff
@@ -5,9 +5,30 @@ pub mod ffi;
...
+
+// Feature validation for incompatible combinations
+#[cfg(all(feature = "fullmerkletree", feature = "optimalmerkletree"))]
+compile_error!("Cannot enable both fullmerkletree and optimalmerkletree");
+
+#[cfg(all(feature = "fullmerkletree", feature = "pmtree-ft"))]
+compile_error!("Cannot enable both fullmerkletree and pmtree-ft");
+
+#[cfg(all(feature = "optimalmerkletree", feature = "pmtree-ft"))]
+compile_error!("Cannot enable both optimalmerkletree and pmtree-ft");
```
seemenkina:
```suggestion
// Ensure that only one Merkle tree feature is enabled at a time
#[cfg(any(
    all(feature = "fullmerkletree", feature = "optimalmerkletree"),
    all(feature = "fullmerkletree", feature = "pmtree-ft"),
    all(feature = "optimalmerkletree", feature = "pmtree-ft"),
))]
compile_error!("Only one of `fullmerkletree`, `optimalmerkletree`, or `pmtree-ft` can be enabled at a time.");
```

rln/src/lib.rs
```diff
@@ -5,9 +5,30 @@ pub mod ffi;
...
+
+#[cfg(all(
+    feature = "stateless",
+    any(
+        feature = "fullmerkletree",
+        feature = "optimalmerkletree",
+        feature = "pmtree-ft"
+    )
+))]
+compile_error!("Cannot enable any merkletree features with stateless");
```
seemenkina:
```suggestion
compile_error!("Cannot enable any Merkle tree features with stateless");
```

rln/src/poseidon_tree.rs
```diff
@@ -20,6 +21,11 @@ cfg_if! {
 
         pub type PoseidonTree = OptimalMerkleTree<PoseidonHash>;
         pub type MerkleProof = OptimalMerkleProof<PoseidonHash>;
+    } else if #[cfg(feature = "pmtree-ft")] {
+        use crate::pm_tree_adapter::{PmTree, PmTreeProof};
+
+        pub type PoseidonTree = PmTree;
+        pub type MerkleProof = PmTreeProof;
```
seemenkina: I think `else` has covered this case

rln/src/protocol.rs
```diff
@@ -240,17 +246,15 @@ pub fn proof_inputs_to_rln_witness(
 /// Returns an error if `message_id` is not within `user_message_limit`.
 pub fn rln_witness_from_values(
     identity_secret: IdSecret,
-    merkle_proof: &MerkleProof,
+    path_elements: Vec<Fp<MontBackend<FrConfig, 4>, 4>>,
+    identity_path_index: Vec<u8>,
```
seemenkina, line 250: why did you replace proof on it parts?

rln/src/poseidon_tree.rs
vinhtc27: Just clarify the `pmtree-ft` feature explicitly if duplication isn't a concern, wdyt?

rln/src/protocol.rs
vinhtc27: This is part of the poseidon_tree.rs file in rln module. The PoseidonTree and MerkleProof types are not meant to be used with the stateless feature. If we want to use them, we need to activate a tree-related feature alongside stateless, as was done previously.
That's why I had to remove the MerkleProof parameter from the rln_witness_from_values function, since this function is used everywhere with the stateless feature.

rln/src/poseidon_tree.rs
seemenkina: I understand what's behind this, but it doesn't make sense.

That is, either we write all the features using `else-if` and then throw an error in `else`, such as:

```
cfg_if::cfg_if! {
    if #[cfg(feature = "fullmerkletree")] {
        // ...
    } else if #[cfg(feature = "optimalmerkletree")] {
        // ...
    } else if #[cfg(feature = "pmtree-ft")] {
        // ...
    } else {
        compile_error!("One of the features `fullmerkletree`, `optimalmerkletree`, or `pmtree-ft` must be enabled.");
    }
}
```

Or we leave it as it was, which, to be honest, makes more sense.

rln/src/protocol.rs
seemenkina: got it, let's leave like this

rln/src/poseidon_tree.rs
vinhtc27: yeah this is better, ill update the code

seemenkina (approved):

PR #330, opened 2025-07-25
feat(rln):  add big endian support for the whole API

seemenkina:
- Prepare the naive solution: when initialising RLN, the required endians is selected and then all RLN methods automatically determine the calculation 
- It is important that BE has not been tested for trees
- ArkProof so far is always serialised and deserialised in LE because libraries provide only such compressed serialisation. In addition on the end-user side this value is not parsed and is stored as bytes.
- And also separated into separate data generation functions to avoid unnecessary initialisation of the rln instance

### DIFF
--- a/rln-wasm/src/lib.rs
+++ b/rln-wasm/src/lib.rs
@@ -135,9 +135,9 @@ impl ProcessArg for bool {
 
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
 #[wasm_bindgen(js_name = newRLN)]
-pub fn wasm_new(zkey: Uint8Array, endianness: bool) -> Result<*mut RLNWrapper, String> {
-    let instance =
-        RLN::new_with_params(zkey.to_vec(), endianness).map_err(|err| format!("{:#?}", err))?;
+pub fn wasm_new(zkey: Uint8Array, is_little_endian: bool) -> Result<*mut RLNWrapper, String> {
+    let instance = RLN::new_with_params(zkey.to_vec(), is_little_endian)
+        .map_err(|err| format!("{:#?}", err))?;
     let wrapper = RLNWrapper { instance };
     Ok(Box::into_raw(Box::new(wrapper)))
 }
--- a/rln-cli/src/examples/stateless.rs
+++ b/rln-cli/src/examples/stateless.rs
@@ -11,7 +11,7 @@ use rln::{
     circuit::{Fr, TEST_TREE_HEIGHT},
     hashers::{hash_to_field_le, poseidon_hash, PoseidonHash},
     protocol::{keygen, prepare_verify_input_le, rln_witness_from_values, serialize_witness_le},
-    public::{Endianness, RLN},
+    public::RLN,
     utils::fr_to_bytes_le,
 };
 use zerokit_utils::{OptimalMerkleTree, ZerokitMerkleProof, ZerokitMerkleTree};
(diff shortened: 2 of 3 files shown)


rln/src/utils.rs
```diff
@@ -3,6 +3,7 @@
 use crate::circuit::Fr;
 use crate::error::ConversionError;
 use ark_ff::PrimeField;
+use ark_std::vec::Vec;
```
sydhds: Any particular reason to use Vec from ark_std instead of Rust std lib?

rln/src/public.rs
```diff
@@ -721,7 +791,14 @@ impl RLN {
     #[cfg(not(feature = "stateless"))]
     pub fn get_empty_leaves_indices<W: Write>(&self, mut output_data: W) -> Result<(), RLNError> {
         let idxs = self.tree.get_empty_leaves_indices();
-        idxs.serialize_compressed(&mut output_data)?;
+        match self.endianness {
+            Endianness::LittleEndian => {
+                idxs.serialize_compressed(&mut output_data)?;
+            }
+            Endianness::BigEndian => {
+                // TODO: implement
```
sydhds: Maybe as a quick impl, you can use serialize_compressed and reverse the byte?

rln/src/public.rs
```diff
@@ -810,9 +890,14 @@ impl RLN {
         // [ proof<128> | root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32> ]
         let mut input_byte: Vec<u8> = Vec::new();
         input_data.read_to_end(&mut input_byte)?;
-        let proof = ArkProof::deserialize_compressed(&mut Cursor::new(&input_byte[..128]))?;
+        // TODO: looks like in nwaku they just store proof as compressed
```
sydhds: TODO here is not very clear of what is missing or what is needed to be impl

rln-wasm/src/lib.rs
```diff
@@ -160,10 +159,24 @@ impl<'a> ProcessArg for &'a [u8] {
...
+    fn process(self) -> Self::ReturnType {
+        self
+    }
+}
+
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
 #[wasm_bindgen(js_name = newRLN)]
-pub fn wasm_new(zkey: Uint8Array) -> Result<*mut RLNWrapper, String> {
-    let instance = RLN::new_with_params(zkey.to_vec()).map_err(|err| format!("{:#?}", err))?;
+pub fn wasm_new(zkey: Uint8Array, endianness: Number) -> Result<*mut RLNWrapper, String> {
```
sydhds: Would it be more explicit to pass a boolean here and so user cannot pass any invalid value here?

sydhds (commented):
All good :) but a few things that you might want to check

rln/src/utils.rs
seemenkina: no reason, just typo

rln/src/public.rs
seemenkina: The point is that Arkproof only supports compressed serialisation in little endian.

However, this data is not converted back to Arkproof on the end user's side, but is simply stored as bytes and then sent for verification. 

So, my hypothesis is that there is no point in changing the conversion and adding, for example, byte reversal for this value

rln-wasm/src/lib.rs
seemenkina: hm, and variable name will be something like (`is_little_endian`)?  Or having name like `endianness` for bool vakue is good enough?

rln/src/public.rs
sydhds: So maybe add this as a comment?

rln-wasm/src/lib.rs
sydhds: is_little_endian looks more explicit I would say

seemenkina: agree, I'll replace it

PR #332, opened 2025-07-30
feat(rln-wasm-utils): extracting the generation and hash functions into a separate module

seemenkina:
- separated all identity generation functions as separate functions, rather than RLN methods
- added BE support - only for these functions so far
- covered the functions with tests, as well as conversion to big endian
- prepared for publication, but is actually awaiting the initial publication of the RLN module

@vinhtc27, please check that everything is correct from the wasm point of view. This module does not require parallel computing, so if there are any unnecessary dependencies, builds, etc., please let me know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
       - "!rln/src/**"
       - "!rln/resources/**"
       - "!utils/src/**"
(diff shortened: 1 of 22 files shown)


vinhtc27 (commented):
I'll try to push the updates

vinhtc27 (approved):

PR #334, opened 2025-08-18
feat(rln): improve the PmTreeConfig initialization process with builder pattern

vinhtc27:


rln/src/pm_tree_adapter.rs
```diff
@@ -50,7 +50,7 @@ impl Hasher for PoseidonHash {
     }
 }
 
-fn get_tmp_path() -> PathBuf {
+pub fn get_tmp_path() -> PathBuf {
```
sydhds: This is a very bad & insecure way to create a temporary file. Better use tempfile (using with_prefix) here no?

rln/src/pm_tree_adapter.rs
```diff
@@ -60,6 +60,89 @@ fn get_tmp() -> bool {
 
 pub struct PmtreeConfig(Config);
 
+pub struct PmtreeConfigBuilder {
```
sydhds, line 69: Does it make sense to make all the values optional? I would assume that either you provide all the values OR you deal with default values.

rln/src/pm_tree_adapter.rs
```diff
@@ -60,6 +60,89 @@ fn get_tmp() -> bool {
...
+    path: Option<PathBuf>,
+    temporary: Option<bool>,
+    cache_capacity: Option<u64>,
+    flush_every_ms: Option<u64>,
+    mode: Option<Mode>,
+    use_compression: Option<bool>,
+}
+
+impl PmtreeConfigBuilder {
```
sydhds, line 78: Cannot you use a crate to create automatically a builder. Maybe using crate: bon or crate: derive_builder ?

rln/src/pm_tree_adapter.rs
```diff
@@ -60,6 +60,89 @@ fn get_tmp() -> bool {
...
+
+    pub fn build(self) -> Result<PmtreeConfig, FromConfigError> {
+        let path = self.path.unwrap_or_else(get_tmp_path);
+        let temporary = self.temporary.unwrap_or_else(get_tmp);
+
+        if temporary && path.exists() {
+            return Err(FromConfigError::PathExists);
+        }
+
+        let config = Config::new()
```
sydhds, line 128: I would define every default values as constant and use the constant later.

rln/src/public.rs
```diff
@@ -66,36 +99,24 @@ impl RLN {
...
     #[cfg(all(not(target_arch = "wasm32"), not(feature = "stateless")))]
-    pub fn new<R: Read>(tree_depth: usize, mut input_data: R) -> Result<RLN, RLNError> {
-        // We read input
-        let mut input: Vec<u8> = Vec::new();
-        input_data.read_to_end(&mut input)?;
-
-        let rln_config: Value = serde_json::from_str(&String::from_utf8(input)?)?;
-        let tree_config = rln_config["tree_config"].to_string();
-
+    pub fn new<T: TreeConfigInput>(tree_depth: usize, input_buffer: T) -> Result<RLN, RLNError> {
```
sydhds, line 115: Why not pass here something like an enum: 

```Rust
enum ConfigOrigin {
    FromRead(Box<dyn Read>),
    FromObj(Option<Cfg>),
}
```

FromRead could even be renamed to FromJsonPath to be more explicit. And you could even add FromJsonString to have more option here.

Then you could implement a standart try_from trait for this enum to get the config. 

Wdyt?

rln/src/pm_tree_adapter.rs
vinhtc27: I tried it, and PmtreeConfig is essentially a wrapper around the Config struct inside the sled crate. However, since it also contains an inner type of Config type, it becomes difficult to go around with derive_builder's macro, which makes the code look confusing.

I also tried creating a fn builder() that deref to the inner builder method of Config (already provided by a macro in the sled crate), but that ends up exposing some unwanted methods to the outside.

So I guess it’s better to stick with a bit of boilerplate code, but at least it stays clean enough for this case.

rln/src/pm_tree_adapter.rs
vinhtc27: Yes, ill make all common value as constant

rln/src/public.rs
vinhtc27: It’s kind of overcomplicated to use an enum here. I think it’s better to keep it simple with a trait in this case.
It’s just two types of input, not an error type that really needs an enum with `Box<dyn ...>` stuff. 
Also, using `dyn` increases compile time compared to `impl` Trait, and `Box` introduces heap allocation.

sydhds: Why is this overcomplicated? User can just do something like

```
let cfg = Cfg::default();
let mut co = ConfigOrigin::FromObj(Some(cfg));
println!("co: {:?}", co);

let tree_depth = 20;

RLN::new(tree_depth, co) {
    let cfg = co.try_into().unwrap();
    // rest of the code
}
```

or with a Read like object:

```
let cfg_read = Cfg::default();
let mut co_r = ConfigOrigin::FromRead(Box::new(std::fs::File::open("myconfig.json").unwrap()));
```

Also: "using dyn increases compile time compared to impl Trait" 

I am not sure this is true. "dyn" is using dynamic dispatch (at runtime) while "impl Trait" is using monomorphization at compile time.

+ "Box introduces heap allocation."

True but only in case of user wanting to pass a dyn Trait object. I would not care here for only 1 possible allocation. And this is not a critical hot path here...

vinhtc27: Thanks for all these insights.
I'd prefer to keep this simple with the trait approach for now. Adding a ConfigOrigin enum would require importing another type across the codebase for just two input cases. We can always refactor later if needed

rln/src/pm_tree_adapter.rs
```diff
@@ -78,18 +163,16 @@ impl FromStr for PmtreeConfig {
...
+            }
         }
 
         let config = Config::new()
-            .temporary(temporary.unwrap_or(get_tmp()))
-            .path(path.unwrap_or(get_tmp_path()))
-            .cache_capacity(cache_capacity.unwrap_or(1024 * 1024 * 1024))
+            .temporary(temporary.unwrap_or(DEFAULT_TEMPORARY))
+            .path(path.unwrap_or(default_tmp_path()))
+            .cache_capacity(cache_capacity.unwrap_or(DEFAULT_CACHE_CAPACITY))
```
seemenkina, line 175: DEFAULT_CACHE_CAPACITY looks smaller than it was previously

vinhtc27: I'l update to previous value

vinhtc27:
The current failing CI is due to this issue with CMake on MacOS.
https://github.com/actions/runner-images/issues/12912

seemenkina (approved):
LGTM

PR #337, opened 2025-09-02
Add initial code for ffi rework

sydhds:


rln/src/protocol.rs
```diff
@@ -37,18 +37,18 @@ use zeroize::Zeroize;
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct RLNWitnessInput {
     #[serde(serialize_with = "ark_se", deserialize_with = "ark_de")]
-    identity_secret: IdSecret,
+    pub identity_secret: IdSecret,
```
seemenkina: why we need to make filed as pub? I don't think that have a straight access to them is a good idea. Better to add getter/setter for required fields

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+        tree,
+    };
+
+    CResult {
+        ok: Some(Box_::new(rln)),
+        err: None,
+    }
+}
+
+#[cfg(feature = "stateless")]
```
seemenkina:
```suggestion
#[cfg(feature = "stateless")]
#[ffi_export]
```

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+        Err(e) => CResult {
+            ok: None,
+            err: Some(e.to_string().into()),
+        },
+    }
+}
+
+#[cfg(not(feature = "stateless"))]
+#[ffi_export]
+pub fn ffi2_generate_rln_proof(
```
seemenkina: So here looks like you do the same thing as in ffi2_generate_rln_proof_with_witness. How it should be: 

`ffi2_generate_rln_proof` - required some of the witness inputs and merkle path we get from local tree instance - see 
`ffi2_generate_rln_proof_with_witness` - required as input whole rln_witness with merkle path 

Perhaps `ffi2_prove` can be removed. I suppose it now fully repeat  `ffi2_generate_rln_proof_with_witness`

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+}
+
+#[ffi_export]
+pub fn ffi2_verify_with_roots(
+    rln: &repr_c::Box<FFI2_RLN>,
+    proof: &repr_c::Box<FFI2_RLNProof>,
+    roots: repr_c::Vec<CFr>,
+) -> CResult<repr_c::Box<bool>, repr_c::String> {
+    // Verify the proof
+    let verified = match verify_proof(&rln.proving_key.0.vk, &proof.proof, &proof.proof_values) {
```
seemenkina: Here you actually missed roots verification, please check how public API works

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+            proof_values,
+            proof,
+        })),
+        err: None,
+    }
+}
+
+#[cfg(not(feature = "stateless"))]
+#[ffi_export]
+pub fn ffi2_verify_rln_proof(
```
seemenkina: Here you actually missed tree instance root verification, please check how public API works

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+        ok: Some(Box_::new(FFI2_RLNProof {
+            proof_values,
+            proof,
+        })),
+        err: None,
+    }
+}
+
+#[ffi_export]
+pub fn ffi2_verify(
```
seemenkina: Regarding this function, I am also considering how necessary it is and whether it has the correct name. The idea was that we had two functions that only work with Groth16 and, during verification, for example, do not check that the root of the tree is correct, but only directly check the zk proof. 

This means that during verification, we can still have three different APIs: 
1) verification of only zk proofs - ?do we need this one for public API?
2) verification of zk proofs + tree proofs (the tree entity is at zerokit and we do not expect it at the input)
3) verification of zk proof + tree proof (all information for stateless mode is required at the input)

@sydhds what do you think? 

So right now all verification function do the same thing, please check it regarding public API

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1019 @@
...
+#[ffi_export]
+fn ffi2_merkle_proof_free(proof: Option<repr_c::Box<FFI2_MerkleProof>>) {
+    drop(proof);
+}
+
+// RLNWitnessInput
+
+#[derive_ReprC]
+#[repr(opaque)]
+pub struct FFI2_RLNWitnessInput {
```
seemenkina: Perhaps we need to do some function for non-stateless mode, because in such case we don't expect path_elements and identity_path_index from users.

rln/src/ffi2.rs
sydhds: @seemenkina I like the idea of having different "verification" function. Just for 2. I am not sure to understand this: "the tree entity is at zerokit and we do not expect it at the input" ?

rln-wasm-utils/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rln = { path = "../rln", version = "0.9.0", default-features = false, features =
 js-sys = "0.3.77"
 wasm-bindgen = "0.2.100"
 rand = "0.8.5"
+ark-bn254 = { version = "0.5.0", features = ["std"] }
```
sydhds: Is this compat with no-parallel feature?

rln/ffi_c_examples/main.c
```diff
@@ -0,0 +1,232 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "rln.h"
+
+int main (int argc, char const * const argv[])
+{
+    printf("Creating RLN instance\n");
+    const char* config = "{"
```
sydhds: Cannot we load a json file here? Would be easier for future updates...

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1018 @@
...
+
+    CResult {
+        ok: Some(Box_::new(rln)),
+        err: None,
+    }
+}
+
+#[cfg(feature = "stateless")]
+#[ffi_export]
+pub fn ffi2_new() -> CResult<repr_c::Box<FFI2_RLN>, repr_c::String> {
```
sydhds: I'm wondering if we could split the file in 2: ffi2.rs (non stateless) && ffi2_stateless.rs for stateless functions. And then include based on the selected features? It would make the file lighter, wdyt?

rln/src/ffi2.rs
seemenkina: > the tree entity is at zerokit and we do not expect it at the input

it is about statefull feature when we don't require Merkle Path and Proof from end user and take this information from local tree

rln/src/ffi2.rs
seemenkina: We have some functions that possible for with/without feature like verification_with_roots

Perphaps we can leave such function in stateless file and all other especially tree FFI move into separate file

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1045 @@
...
+    }
+}
+
+////////////////////////////////////////////////////////
+// zkSNARKs APIs
+////////////////////////////////////////////////////////
+
+#[cfg(not(feature = "stateless"))]
+#[ffi_export]
+pub fn ffi2_generate_rln_proof(
```
seemenkina: This one actually the whole copy of ffi2_generate_rln_proof_with_witness

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1045 @@
...
+#[ffi_export]
+pub fn ffi2_verify_rln_proof(
+    rln: &repr_c::Box<FFI2_RLN>,
+    proof: &repr_c::Box<FFI2_RLNProof>,
+) -> CResult<repr_c::Box<bool>, repr_c::String> {
+    // Verify the proof
+    match verify_proof(&rln.proving_key.0.vk, &proof.proof, &proof.proof_values) {
+        Ok(proof_verified) => {
+            // Verify the root
+            let roots_verified = rln.tree.root() == proof.proof_values.root;
```
seemenkina: So where is the message itself? Your input should be proof_data and signal and then you also do check, that proof_data.x == signal

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1045 @@
...
+            Err(err) => {
+                return CResult {
+                    ok: None,
+                    err: Some(err.to_string().into()),
+                };
+            }
+        };
+
+    // If proof verification failed, return early
+    if !proof_verified {
```
seemenkina: Same here, you need to check that we actually have a proof for our signal

rln/src/protocol.rs
seemenkina: just in case, please do not to close comment if you didn't provide answer or changes. 

So, here it's a bad practice to make straight access to the fields. Let's do some builder pattern with sanity checks like path_elements and identity_path_index has the same length, message_id range check etc

rln/ffi_c_examples/main.c
seemenkina: Agree, looks a bit difficult to maintain

seemenkina:
image

rln-wasm-utils/Cargo.toml
vinhtc27: What do you mean by the no-parallel feature? There isn’t a parallel feature in ark-bn254.

rln/src/ffi2.rs
seemenkina: Pleas move this one next to ffi2_generate_rln_proof_statelss

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1072 @@
...
+
+    let proof_values = match proof_values_from_witness(&rln_witness) {
+        Ok(pv) => pv,
+        Err(err) => {
+            return CResult {
+                ok: None,
+                err: Some(err.to_string().into()),
+            };
+        }
+    };
```
seemenkina: What do you think about move this part inside generate_proof function? Because we call it every time before generation

rln/src/ffi2.rs
```diff
@@ -0,0 +1,1072 @@
...
+pub fn ffi2_hash_to_field_le(input: c_slice::Ref<'_, u8>) -> repr_c::Box<CFr> {
+    let hash_result = hash_to_field_le(&input);
+    CFr::from(hash_result).into()
+}
+
+#[ffi_export]
+pub fn ffi2_hash_to_field_be(input: c_slice::Ref<'_, u8>) -> repr_c::Box<CFr> {
+    let hash_result = hash_to_field_be(&input);
+    CFr::from(hash_result).into()
+}
```
seemenkina: So maybe here we return to one hash function as we now have to_bytes_le/to_bytes_be

rln/src/ffi2.rs
seemenkina: So @vinhtc27 let's do it next way: 
Let's create ffi folder in src and inside this folder create next files:  
- ffi_utils.rs - for all utils function + structure defintion
- ffi_rln.rs - here only new functions, proof generation functions, proof verifications function 
- ffi_tree.rs - here all tree related functions

@sydhds wdyt? I'm not sure that stateless/non-sateless make sense as we have for example verify_with_roots which is work for both

rln/tests/ffi2.rs
```diff
@@ -0,0 +1,1701 @@
+#[cfg(test)]
+#[cfg(not(feature = "stateless"))]
```
seemenkina: Let's make two files with stateless tests and non statless test. This one is quite big

rln/src/ffi2.rs
vinhtc27: You mean merge two functions into one, with an additional boolean argument for endianness?

rln/src/ffi2.rs
vinhtc27: We still need to return the `proof_values` along with the `proof`. 
Unless we create a type that contains both, we should probably just keep this outside. 
After refactoring all the public APIs later on, it’s only used in four places anyway.

seemenkina (approved):
LGTM, thank you!

PR #342, opened 2025-09-14
fix(rln): fixed fail nightly build and updated CONTRIBUTING.md and Cargo.lock

vinhtc27:
### Update Cargo.lock
+ Clean lock file.

### Update CONTRIBUTING.md 
+ Fixed the clippy command to avoid compilation errors.
+ Added the rln-wasm-utils module.

### Fixed nightly build failures
When building with the following feature combinations:
+ fullmerkletree
+ fullmerkletree,parallel
+ optimalmerkletree
+ optimalmerkletree,parallel

Failed CI run (master): https://github.com/vacp2p/zerokit/actions/runs/17707099741

+ Since rln-cli activates pmtree-ft by default, we need to exclude rln-cli from the workspace.
+ Also, when implementing the new TreeConfigInput trait for <PoseidonTree as ZerokitMerkleTree>::Config, it must remain behind the pmtree-ft feature flag to avoid conflicts with other tree config types.

Successful CI run (this PR branch): https://github.com/vacp2p/zerokit/actions/runs/18092958207


CONTRIBUTING.md
```diff
@@ -105,8 +105,9 @@ docs: update RLN protocol flow documentation
 Use scopes to improve the Changelog:
 
 - `rln` - Core RLN implementation
-- `rln-wasm` - WebAssembly bindings
 - `rln-cli` - Command-line interface
+- `rln-wasm` - WebAssembly bindings
+- `rln-wasm-utils` - WebAssembly utilities
```
sydhds, line 110: utilities? If I recall correctly, this is another wasm file with only the keygen related functons right?

vinhtc27: Yeah, for now it is.
I guess it’s intended to move all utility functions for WASM here, and possibly more in the future. 
WDYT? Any naming suggestions here or should I just remove this line?

seemenkina (approved):
LGTM

CONTRIBUTING.md
seemenkina: Yes, the idea is to move all auxiliary functions to `rln-wasm-utils`, since most of them are not actually related to rln, but are used by rln, so perhaps a different name would be more appropriate

PR #352, opened 2025-10-31
feat(wasm): rework rln-wasm and rln-wasm-utils modules, remove buffer-based serialization, and update public.rs and protocol.rs accordingly

vinhtc27:


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rust-analyzer.linkedProjects": [
+    "./rln/Cargo.toml",
+    "./rln-cli/Cargo.toml",
+    "./rln-wasm/Cargo.toml",
+    "./utils/Cargo.toml"
+  ]
+}
```
seemenkina: please, add this file into ignore, we don't need it in repo

rln-wasm/src/wasm_rln.rs
```diff
@@ -0,0 +1,191 @@
...
+        WasmFr::from(self.proof_values.x)
+    }
+
+    #[wasm_bindgen(getter, js_name = externalNullifier)]
+    pub fn external_nullifier(&self) -> WasmFr {
+        WasmFr::from(self.proof_values.external_nullifier)
+    }
+
+    #[wasm_bindgen(js_name = recoverIdSecret)]
+    pub fn recover_id_secret(
```
seemenkina, line 119: no need to make this function as method as it not use self, but required to create entity

rln-wasm/Cargo.toml
```diff
@@ -34,14 +37,17 @@ getrandom = { version = "0.2.16", features = ["js"] }
...
 
 [dev-dependencies.web-sys]
 version = "0.3.77"
 features = ["Window", "Navigator"]
 
 [features]
-default = ["console_error_panic_hook"]
-parallel = ["rln/parallel", "wasm-bindgen-rayon"]
+default = []
+utils = []
```
seemenkina, line 48: not sure that I agree about such separation as it not obvious in code especially how to build prover+utils and just utils 

@sydhds wdyth

rln-wasm/src/wasm_rln.rs
```diff
@@ -0,0 +1,191 @@
...
+        rln_witness_to_bigint_json, verify_proof, RLNProofValues, RLNWitnessInput,
+    },
+    utils::IdSecret,
+};
+use serde::Serialize;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub struct WasmRLN {
+    proving_key: (ProvingKey<Curve>, ConstraintMatrices<Fr>),
```
seemenkina: Maybe we should move this part into the RLN protocol itself. It feels too specific for the FFI layer, so it would be better to keep this logic under an abstraction.

I feel the same way about RLNProof.

My idea is that, whenever possible, we should avoid type specifications in WASM/FFI. That way, if there are changes to the zk-proof, only the RLN core needs to be updated—not the WASM/FFI interface.

rln-wasm/src/wasm_rln.rs
vinhtc27: You can just call `rlnWasm.WasmRLNProof.recoverIdSecret()` directly, so there's no need to create an entity.  
Same for `Uint8ArrayUtils`, for example use `rlnWasm.Uint8ArrayUtils.toBytesLE()` in Javascript.  

The object types help make things clearer, so keeping them inline makes the code easier to read.

rln-wasm/src/wasm_rln.rs
vinhtc27: Agree, I’ll do the `protocol.rs` refactoring directly in this PR then.

rln-wasm/Cargo.toml
vinhtc27: It’s kinda cleaner this way. I follow the same stateless and non-stateless pattern here for the less feature flags

seemenkina (approved):
LGTM, let's finish this iteration

PR #354, opened 2025-11-11
fix(rln): resolve memory leak in calc_witness and improve FFI memory deallocation pattern

vinhtc27:
### Memory Leak Investigation

Using `valgrind` and `leaks`, I identified the following memory leak issues:

- The call to `unsafe { zeroize_flat_type(v) }` in the `calc_witness` function caused a memory leak.  
  → ~I tried assigning `U256::ZERO` instead, though it’s unclear if the compiler will optimize it away.~  
  → This happened because the current code zeroized the Vec pointer, not the data itself.  
  **Fix:** Iterate over each `U256` element and free it individually with `zeroize_flat_type`.

- The `Box<T>` wrappers for the `CResult<T: ReprC, Err: ReprC>` types were not freed in the C and Nim examples.  
  → ~All `T` types returned via `CResult` are boxed for `NULL` checking and should have their `.ok` value freed afterward.~  
  **Fix:** Modified the return type so it can be freed directly via a Rust-defined function, simplifying both error handling and logging. `CResult<Box<bool>, CString>` type is simplify into `Option<CString>` to represent error if != NULL.

### DIFF
--- a/rln/src/ffi/ffi_utils.rs
+++ b/rln/src/ffi/ffi_utils.rs
@@ -129,17 +129,13 @@ pub fn vec_cfr_from_cfr(cfr: &CFr) -> repr_c::Vec<CFr> {
 }
 
 #[ffi_export]
-pub fn vec_cfr_push(v: &mut repr_c::Vec<CFr>, cfr: &CFr) {
-    let ptr = v.as_mut_ptr();
-    let len = v.len();
-
-    let cap = unsafe { std::ptr::read((v as *const _ as *const usize).add(2)) };
-
-    let mut rust_vec = unsafe { Vec::from_raw_parts(ptr, len, cap) };
-    rust_vec.push(*cfr);
-
-    let new_repr_vec: repr_c::Vec<CFr> = rust_vec.into();
-    unsafe { std::ptr::write(v, new_repr_vec) };
+pub fn vec_cfr_push(v: &mut safer_ffi::Vec<CFr>, cfr: &CFr) {
+    let mut new: Vec<CFr> = std::mem::replace(v, Vec::new().into()).into();
+    if new.len() == new.capacity() {
+        new.reserve_exact(1);
+    }
+    new.push(*cfr);
+    *v = new.into();
 }
 
 #[ffi_export]


rln/src/circuit/iden3calc.rs
```diff
@@ -43,12 +42,11 @@ pub fn calc_witness<I: IntoIterator<Item = (String, Vec<FrOrSecret>)>>(
     let mut inputs_buffer = get_inputs_buffer(get_inputs_size(&nodes));
     populate_inputs(&inputs, &input_mapping, &mut inputs_buffer);
     if let Some(v) = inputs.get_mut("identitySecret") {
-        // ~== v[0] = U256::ZERO;
-        unsafe { zeroize_flat_type(v) };
+        v[0] = U256::ZERO;
```
sydhds: This is not a valid "Zeroize" call. 

I would advise here: 
* Try update to zeroize 1.8.2
* If not, try to repro the error with a minimal example, and report a bug to zeroize crate

rln/ffi_c_examples/main.c
```diff
@@ -16,7 +16,8 @@ int main (int argc, char const * const argv[])
 #endif
 
     if (!ffi_new_result.ok) {
```
sydhds: Naming is a bit weird here, maybe we should have something like: `_TYPE_ ffi_rln_result = ffi_rln_new();` ?

rln/ffi_c_examples/main.c
```diff
@@ -76,19 +77,23 @@ int main (int argc, char const * const argv[])
 
     CResult_Vec_CFr_ptr_Vec_uint8_t deser_keys_result = bytes_le_to_vec_cfr(&ser_keys);
```
sydhds: type here is way too long. Coud we have more meaningful names here?

rln/ffi_c_examples/main.c
```diff
@@ -255,134 +271,140 @@ int main (int argc, char const * const argv[])
...
-        printf("\nVerifying Proof\n");
+    printf("\nVerifying Proof\n");
 #ifdef STATELESS
-        Vec_CFr_t roots = {
-            .ptr = computed_root,
-            .len = 1,
-            .cap = 1
-        };
-        CResult_bool_ptr_Vec_uint8_t verify_result = ffi_verify_with_roots(&rln, &rln_proof, &roots, x);
+    Vec_CFr_t roots = {
```
sydhds: Maybe add function to init a Vec_CFr_t from a CFr ?

vinhtc27: It's not possible to create a type and then push to it like a `Vec` in Rust. I tried it.

rln/src/circuit/iden3calc.rs
vinhtc27: The code zeroized the Vec pointer instead of the data. I found a fix for it.

rln/src/ffi/ffi_utils.rs
```diff
@@ -162,25 +165,31 @@ pub fn bytes_be_to_vec_cfr(
...
-    crate::utils::vec_u8_to_bytes_le(vec).into()
+pub fn vec_u8_to_bytes_le(vec: &repr_c::Vec<u8>) -> repr_c::Box<repr_c::Vec<u8>> {
+    Box_::new(crate::utils::vec_u8_to_bytes_le(vec).into())
 }
 
 #[ffi_export]
-pub fn vec_u8_to_bytes_be(vec: &repr_c::Vec<u8>) -> repr_c::Vec<u8> {
-    crate::utils::vec_u8_to_bytes_be(vec).into()
+pub fn vec_u8_to_bytes_be(vec: &repr_c::Vec<u8>) -> repr_c::Box<repr_c::Vec<u8>> {
+    Box_::new(crate::utils::vec_u8_to_bytes_be(vec).into())
```
sydhds: Unclear what is done here (need docstring?). a Vec<u8> is turned into another Vec<u8>?

rln/ffi_c_examples/main.c
```diff
@@ -4,177 +4,194 @@
...
+int main(int argc, char const *const argv[])
 {
     printf("Creating RLN instance\n");
 
 #ifdef STATELESS
     CResult_FFI_RLN_ptr_Vec_uint8_t ffi_new_result = ffi_new();
 #else
-    const char* config_path = "../resources/tree_depth_20/config.json";
+    const char *config_path = "../resources/tree_depth_20/config.json";
     CResult_FFI_RLN_ptr_Vec_uint8_t ffi_new_result = ffi_new(20, config_path);
```
sydhds: ffi_new -> ffi_rln_new?

rln/ffi_c_examples/main.c
sydhds: Can you elaborate on this? Here we are talking about creating a Vec_Cfr_t from a Cfr_t so not sure to understand why it cannot be done here...

rln/src/circuit/iden3calc.rs
```diff
@@ -41,15 +41,21 @@ pub fn calc_witness<I: IntoIterator<Item = (String, Vec<FrOrSecret>)>>(
         deserialize_witnesscalc_graph(std::io::Cursor::new(graph_data)).unwrap();
 
     let mut inputs_buffer = get_inputs_buffer(get_inputs_size(&nodes));
+
     populate_inputs(&inputs, &input_mapping, &mut inputs_buffer);
+
     if let Some(v) = inputs.get_mut("identitySecret") {
-        // ~== v[0] = U256::ZERO;
-        unsafe { zeroize_flat_type(v) };
+        for val in v.iter_mut() {
```
sydhds, line 50: Add a comment why we cannot call it on v directly (for future readers)

rln/src/ffi/ffi_rln.rs
```diff
@@ -364,25 +364,31 @@ pub fn ffi_verify_rln_proof(
     rln: &repr_c::Box<FFI_RLN>,
     proof: &repr_c::Box<FFI_RLNProof>,
     x: &CFr,
-) -> CResult<repr_c::Box<bool>, repr_c::String> {
+) -> Option<repr_c::String> {
```
sydhds: Returning Option<ErrString> is weird here. Can we just return CResult<bool, reprc::String>. I think there would be no need to Box for ok variant as bool is just a primitive type..

rln/src/circuit/iden3calc.rs
```diff
@@ -41,15 +41,24 @@ pub fn calc_witness<I: IntoIterator<Item = (String, Vec<FrOrSecret>)>>(
         deserialize_witnesscalc_graph(std::io::Cursor::new(graph_data)).unwrap();
 
     let mut inputs_buffer = get_inputs_buffer(get_inputs_size(&nodes));
+
     populate_inputs(&inputs, &input_mapping, &mut inputs_buffer);
+
     if let Some(v) = inputs.get_mut("identitySecret") {
-        // ~== v[0] = U256::ZERO;
-        unsafe { zeroize_flat_type(v) };
+        // unsafe { zeroize_flat_type(v) };
```
sydhds: Remove this line and adjust the comment please

rln/src/ffi/ffi_rln.rs
vinhtc27: The `CResult<bool, reprc::String>` version doesn’t work, it won’t allow to return any basic type without boxing it.

I tried changing:

```rust
#[derive_ReprC]
#[repr(C)]
pub struct CResult<T: Any, Err: ReprC> {
    pub ok: Option<T>,
    pub err: Option<Err>,
}
```
but it still won’t accept any T type unless it’s boxed. 
The derive_ReprC macro doesn’t support arbitrary types, only those in repr_c::*.

That’s why I switched to using Option, to avoid manually freeing a Box<bool> on the C side.

```text
error[E0277]: the trait bound `CResult_Layout<bool, safer_ffi::String>: CType` is not satisfied
  --> rln/src/ffi/ffi_tree.rs:35:6
   |
35 | ) -> CResult<bool, repr_c::String> {
   |      ^^^^^^^ unsatisfied trait bound
   |
help: the trait `CType` is not implemented for `CResult_Layout<bool, safer_ffi::String>`
  --> rln/src/ffi/ffi_utils.rs:17:1
   |
15 | #[derive_ReprC]
   | --------------- in this attribute macro expansion
16 | #[repr(C)]
17 | pub struct CResult<T: Any, Err: ReprC> {
   | ^^^
   = help: the trait `CType` is implemented for `CResult_Layout<T, Err>`
   = note: required for `CResult<bool, safer_ffi::String>` to implement `ConcreteReprC`
   = note: this error originates in the attribute macro `derive_ReprC` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #355, opened 2025-11-25
feat: unify RLN types, refactor public APIs, add full (de)serialization, align FFI/WASM/APIs, simplify errors, update docs/examples, and clean up zerokit

vinhtc27:
# Changes

- Unified the `RLN` struct and core protocol types across public, FFI, and WASM so everything works consistently.
- Fully refactored `protocol.rs` and `public.rs` to clean up the API surface and make the flow easier to work with.
- Added (de)serialization for `RLN_Proof` and `RLN_ProofValues`, and matched all C, Nim, WASM, and Node.js examples.
- Aligned FFI and WASM behavior, added missing APIs, and standardized how witness are created and passed around.
- Reworked the error types, added clearer verification messages, and simplified the overall error structure.
- Updated variable names, README, Rust docs, and examples across the repo, updated outdated RLN RFC link.
- Refactored `rln-cli` to use the new public API, removed serialize-based cli example, and dropped the `eyre` crate.
- Bumped dependencies, fixed CI, fixed `+atomic` flags for latest nightly Rust and added `Clippy.toml` for better fmt.
- Added a `prelude.rs` file for easier use, cleaned up public access for types and types import across zerokit modules.
- Separated keygen, proof handling, slashing logic, and witness into protocol folder
- Finished a full review and cleaned up any dust left to start the PR review.


rln-wasm/src/wasm_rln.rs
```diff
@@ -1,42 +1,31 @@
...
 use num_bigint::BigInt;
-use rln::{
-    circuit::{zkey_from_raw, Fr, Proof},
-    protocol::{
-        compute_id_secret, generate_proof_with_witness, proof_values_from_witness,
-        rln_witness_to_bigint_json, verify_proof, RLNProofValues, RLNWitnessInput, RLN,
-    },
-    utils::IdSecret,
-};
+use rln::prelude::*;
```
seemenkina, line 6: Let's I point only here, but I supposed that you go trough all codebase as well.

So using prelude with re-export is a cool idea, but it doesn't mean that we now use glob(*) important everywhere

Why?
- Glob imports bring many names into scope, increasing collision risk
- If a prelude adds a new export that conflicts with end-user code, glob imports can break silently

rln/src/protocol.rs
```diff
@@ -1,61 +1,43 @@
 // This crate collects all the underlying primitives used to implement RLN
```
seemenkina: I have an idea how to make this part more readable. Let's create protocol dir (`rln/src/protocol/`) with next files:

### types.rs 
- RLNWitnessInput struct + impl block
- RLNProof struct
- RLNProofValues struct

### serialization.rs   
All serialization/deserialization functions:
**Witness serialization:**
- rln_witness_to_bytes_le/be
- bytes_le/be_to_rln_witness
- rln_witness_to_bigint_json

 **Proof values serialization:**
- rln_proof_values_to_bytes_le/be
- bytes_le/be_to_rln_proof_values

 **Proof serialization:**
- rln_proof_to_bytes_le/be
- bytes_le/be_to_rln_proof

### keygen.rs
- keygen()
- extended_keygen()
- seeded_keygen()
- extended_seeded_keygen()

### slashing.rs (not sure if this name is good)
- compute_id_secret() 
- recover_id_secret()

### proof.rs 
- calculated_witness_to_field_elements() 
- inputs_for_witness_calculation()
- generate_proof_with_witness()
- generate_proof() 
- verify_proof()
- proof_values_from_witness()
- compute_tree_root() 

### mod.rs
- re-export and all general staff

rln/src/protocol.rs
```diff
@@ -592,34 +649,26 @@ fn calculate_witness_element<E: ark_ec::pairing::Pairing>(
...
         };
-        witness_vec.push(E::ScalarField::from(w))
+        field_elements.push(E::ScalarField::from(w))
     }
 
-    Ok(witness_vec)
+    Ok(field_elements)
 }
 
+/// Generates a RLN proof from pre-calculated witness values.
```
seemenkina: Here we generate zk-proof, not rln-proof. Maybe we also need to rename function like `generate_zk_proof_with_witness`

rln/src/protocol.rs
```diff
@@ -673,44 +712,28 @@ pub fn inputs_for_witness_calculation(
...
+        ("externalNullifier", vec![witness.external_nullifier.into()]),
     ])
 }
 
-/// Generates a RLN proof
-///
-/// # Errors
-///
-/// Returns a [`ProofError`] if proving fails.
+/// Generates a RLN proof from witness input using the provided circuit data.
```
seemenkina: Same here, let's rename to `generate_zk_proof` and update comment

rln/src/protocol.rs
```diff
@@ -721,23 +744,18 @@ pub fn generate_proof(
...
-/// Verifies a given RLN proof
+/// Verifies a RLN proof against the verifying key and public values.
 ///
-/// # Errors
-///
-/// Returns a [`ProofError`] if verifying fails. Verification failure does not
-/// necessarily mean the proof is incorrect.
+/// Returns `true` if the proof is cryptographically valid, `false` if verification fails.
+/// Note: Verification failure may occur due to proof computation errors, not necessarily malicious proofs.
 pub fn verify_proof(
```
seemenkina: Let's also rename to `verify_zk_proof`

rln/src/public.rs
```diff
@@ -598,117 +460,62 @@ impl RLN {
...
-    /// rln.get_proof(index, &mut buffer).unwrap();
-    ///
-    /// let buffer_inner = buffer.into_inner();
-    /// let (path_elements, read) = bytes_le_to_vec_fr(&buffer_inner);
-    /// let (identity_path_index, _) = bytes_le_to_vec_u8(&buffer_inner[read..].to_vec());
+    /// let (path_elements, identity_path_index) = rln.get_proof(index).unwrap();
     /// ```
     #[cfg(not(feature = "stateless"))]
-    pub fn get_proof<W: Write>(&self, index: usize, mut output_data: W) -> Result<(), RLNError> {
+    pub fn get_proof(&self, index: usize) -> Result<(Vec<Fr>, Vec<u8>), RLNError> {
```
seemenkina: let's rename to `get_merkle_proof`

rln/src/public.rs
```diff
@@ -722,826 +529,155 @@ impl RLN {
...
-        input_data.read_to_end(&mut serialized_witness)?;
-        let (rln_witness, _) = deserialize_witness(&serialized_witness)?;
-
-        let proof = generate_proof(&self.zkey, &rln_witness, &self.graph_data)?;
-
-        // Note: we export a serialization of ark-groth16::Proof not semaphore::Proof
-        proof.serialize_compressed(&mut output_data)?;
-
-        Ok(())
+    pub fn generate_proof(&self, witness: &RLNWitnessInput) -> Result<Proof, RLNError> {
```
seemenkina: as well `generate_zk_proof`

rln/src/public.rs
```diff
@@ -722,826 +529,155 @@ impl RLN {
...
+    #[cfg(not(feature = "stateless"))]
+    pub fn flush(&mut self) -> Result<(), ZerokitMerkleTreeError> {
+        self.tree.close_db_connection()
     }
 
-    ////////////////////////////////////////////////////////
     // zkSNARK APIs
-    ////////////////////////////////////////////////////////
+
     /// Computes a zkSNARK RLN proof using a [`RLNWitnessInput`](crate::protocol::RLNWitnessInput) object.
```
seemenkina:
```suggestion
    /// Generates a zkSNARK proof component of an RLN proof from a [`RLNWitnessInput`](crate::protocol::RLNWitnessInput).
```

rln/src/public.rs
```diff
@@ -722,826 +529,155 @@ impl RLN {
...
-        all_read += 8;
-
-        let signal: Vec<u8> = serialized[all_read..all_read + signal_len].to_vec();
-
-        let verified = verify_proof(&self.verifying_key, &proof, &proof_values)?;
-        let x = hash_to_field_le(&signal);
-
-        // Consistency checks to counter proof tampering
-        Ok(verified && (self.tree.root() == proof_values.root) && (x == proof_values.x))
+    pub fn verify_proof(
```
seemenkina:
```suggestion
    pub fn verify_zk_proof(
```

rln/README.md
seemenkina: let's here fix to version, not link to the libarary

rln/src/utils.rs
seemenkina, line 60: So we have checks when we convert arrays to bytes, but we don't have such checks here.  if `input.len() < el_size` we can have runtime panic on prod, so let's also return here result

rln/src/pm_tree_adapter.rs
seemenkina:
```suggestion
                .ok_or(ZerokitMerkleTreeError::InvalidSubTreeIndex)?;
```

rln/src/protocol.rs
```diff
@@ -391,117 +362,182 @@ pub fn deserialize_proof_values(serialized: &[u8]) -> (RLNProofValues, usize) {
...
-    message_id: Fr,
-    external_nullifier: Fr,
-    signal: &[u8],
-) -> Vec<u8> {
+/// Deserializes RLN proof values from big-endian bytes.
+///
+/// Format: `[ root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32> ]`
+///
+/// Returns the deserialized proof values and the number of bytes read.
+pub fn bytes_be_to_rln_proof_values(bytes: &[u8]) -> (RLNProofValues, usize) {
```
seemenkina: So here we also will return result when we add result to `bytes_be_to_fr`

rln/src/ffi/ffi_rln.rs
```diff
@@ -33,158 +22,89 @@ pub fn ffi_rln_new(
...
-        tree,
-    };
-
-    CResult {
-        ok: Some(Box_::new(FFI_RLN(rln))),
-        err: None,
+    let config_str = File::open(config_path.to_str())
+        .and_then(|mut file| {
+            let mut s = String::new();
+            file.read_to_string(&mut s)?;
```
seemenkina, line 42: let's add some limit for reading to avoid buffer overflow. I think 1MB for config is enough?

rln/src/utils.rs
seemenkina: Can we make const from this number to emphasis what and how we calculate?

rln/src/public.rs
```diff
@@ -598,117 +460,62 @@ impl RLN {
...
-    ///
-    /// let buffer_inner = buffer.into_inner();
-    /// let (path_elements, read) = bytes_le_to_vec_fr(&buffer_inner);
-    /// let (identity_path_index, _) = bytes_le_to_vec_u8(&buffer_inner[read..].to_vec());
+    /// let (path_elements, identity_path_index) = rln.get_proof(index).unwrap();
     /// ```
     #[cfg(not(feature = "stateless"))]
-    pub fn get_proof<W: Write>(&self, index: usize, mut output_data: W) -> Result<(), RLNError> {
+    pub fn get_proof(&self, index: usize) -> Result<(Vec<Fr>, Vec<u8>), RLNError> {
         let merkle_proof = self.tree.proof(index).expect("proof should exist");
```
seemenkina: Let's here use map_err, not expect

rln-wasm/src/wasm_rln.rs
vinhtc27: Hey, thanks for catching this. The codebase is following this pattern:
+ Internal modules (protocol.rs, public.rs, utils.rs, etc.) → explicit imports only
+ FFI/WASM wrappers and external consumers (tests, examples) → prelude::*

So glob imports are not re-imported internally. I did try to use them once, but the compiler prevented me from doing so inside the rln module.

As for the user-facing side and examples,  remove the suggestion to use ::* is a valid point. 

I’ll update the examples in rln-cli and the README to avoid that.

rln/src/protocol.rs
vinhtc27: This feels a bit fragmented to me. 
My idea was to move all the core logic and sanity checks into protocol.rs and utils.rs, so that the FFI and WASM layers can mirror that exact structure. Their job would simply be to wrap the types and follow the same ordering:
- the RLN type system
- the proof/verify functions
- the serialization utilities for all types

The public API would then be a thin wrapper around protocol.rs plus the Merkle tree, and both the FFI and WASM layers would reuse that same structure.

seemenkina: @vinhtc27 you make good point. Maybe I split things too much—separating types and serialization is not ideal.

But to be honest, I feel that putting all core logic into one protocol.rs  will end up with giant 'God file' that is nightmare to maintain.

Let us meet in middle. We can structure by feature. We keep the data, the logic, and the serialization together in same file. This is very standard Rust way.

rln/src/protocol/
- mod.rs          // Re-exports everything (the "thin wrapper" you want)
- witness.rs      // Witness Struct + Logic + Serialization
- proof.rs        // Proof Struct + Verify + Serialization
- keygen.rs         // Keygen logic
- slashing.rs     // ID Secret recovery

This way, mod.rs gives clean, flat API for FFI and WASM, but inside, code stays organized and we don't have file with a lot of lines. wdyt?

cc: @sydhds

vinhtc27: That sounds perfect. The file separation kinda matches the flow of the RLN too. 
I'll refactor it this way.

seemenkina (approved):
LGTM, thank you! Great work!

PR #358, opened 2025-12-09
feat(rln): extend error handling for rln module

vinhtc27:
Changes:
- Unified error types (`PoseidonError`, `HashError`, etc.) across hashing, keygen, witness calculation, and serialization for consistent and descriptive error handling.
- Refactored tests and examples to use `unwrap()` where safe, and limited `expect()` in library code to non-panicking cases with clear messaging.
- Improved witness and proof generation by removing panicking code paths and enforcing proper error propagation.
- Cleaned up outdated imports, removed unused operations in `graph.rs`, and updated public API documentation.
- Updated C, Nim, and WASM FFI bindings with more robust serialization and clearer error log messages.
- Added keywords to package.json and update dependencies in Makefile.toml and Nightly CI.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -129,6 +129,9 @@ jobs:
             sed -i.bak 's/rln-wasm/zerokit-rln-wasm/g' pkg/package.json && rm pkg/package.json.bak
           fi
 
+          jq '. + {keywords: ["zerokit", "rln", "wasm"]}' pkg/package.json > pkg/package.json.tmp && \
+          mv pkg/package.json.tmp pkg/package.json
+
           mkdir release
           cp -r pkg/* release/
           tar -czvf rln-wasm-${{ matrix.feature }}.tar.gz release/
--- a/rln-wasm/Makefile.toml
+++ b/rln-wasm/Makefile.toml
@@ -1,14 +1,14 @@
 [tasks.build]
 clear = true
-dependencies = ["pack_build", "pack_rename"]
+dependencies = ["pack_build", "pack_rename", "pack_add_keywords"]
 
 [tasks.build_parallel]
 clear = true
-dependencies = ["pack_build_parallel", "pack_rename"]
+dependencies = ["pack_build_parallel", "pack_rename", "pack_add_keywords"]
 
 [tasks.build_utils]
 clear = true
-dependencies = ["pack_build_utils", "pack_rename_utils"]
+dependencies = ["pack_build_utils", "pack_rename_utils", "pack_add_keywords"]
 
 [tasks.pack_build]
 command = "wasm-pack"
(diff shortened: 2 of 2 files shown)


seemenkina:
@vinhtc27 could you please also change expect to proper error handle? Because expect return panic if something going wrong

rln/src/utils.rs
```diff
@@ -283,12 +283,16 @@ pub fn bytes_le_to_vec_usize(input: &[u8]) -> Result<Vec<usize>, UtilsError> {
                 actual: input.len(),
             });
         }
-        let elements: Vec<usize> = input[8..]
+        input[8..]
             .chunks(8)
```
seemenkina: you can use chunks_exact(8) here:

```
input[8..]
  .chunks_exact(8)
  .take(nof_elem)
  .map(|ch| {
     ch.try_into()
        .map(usize::from_le_bytes)
        .map_err(UtilsError::FromSlice)
   })
   .collect()
```

rln/src/utils.rs
```diff
@@ -310,12 +314,16 @@ pub fn bytes_be_to_vec_usize(input: &[u8]) -> Result<Vec<usize>, UtilsError> {
...
```
seemenkina: and here also please use chunks_exact

rln/src/prelude.rs
```diff
@@ -1,7 +1,7 @@
 // This module re-exports the most commonly used types and functions from the RLN library
 
 #[cfg(not(feature = "stateless"))]
-pub use utils::{Hasher, ZerokitMerkleProof, ZerokitMerkleTree};
+pub use utils::*;
```
seemenkina: not sure if it's a good idea as it becomes less control in future

rln/src/protocol/mod.rs
```diff
@@ -1,9 +1,9 @@
 // This crate collects all the underlying primitives used to implement RLN
 
-mod keygen;
-mod proof;
-mod slashing;
-mod witness;
+pub mod keygen;
```
seemenkina: do we really need make it public?

rln/src/circuit/iden3calc/graph.rs
```diff
@@ -39,13 +34,13 @@ where
...
-pub fn fr_to_u256(x: &Fr) -> U256 {
+pub(crate) fn fr_to_u256(x: &Fr) -> U256 {
     U256::from_limbs(x.into_bigint().0)
 }
 
 #[inline(always)]
-pub fn u256_to_fr(x: &U256) -> Fr {
-    Fr::from_bigint(BigInt::new(x.into_limbs())).expect("Failed to convert U256 to Fr")
+pub(crate) fn u256_to_fr(x: &U256) -> Fr {
+    Fr::from_bigint(BigInt::new(x.into_limbs())).expect("U256 to Fr conversion must be valid")
```
seemenkina: so why you leave expect here? looks like all this function ended in evaluate() -> calc_witness() which return proper error?

rln/src/circuit/iden3calc/graph.rs
```diff
@@ -212,37 +171,27 @@ impl From<&Operation> for proto::DuoOp {
...
-                } else {
-                    M - a
-                }
-            }
-            UnoOperation::Id => a,
-        }
-    }
-
-    pub fn eval_fr(&self, a: Fr) -> Fr {
+    fn eval_fr(&self, a: Fr) -> Result<Fr, String> {
```
seemenkina, line 180: why do you return string and not WitnessCalcError?

vinhtc27: It have multiple other error within this file and i later on map it into WitnessCalcError::GraphEvaluation(String) error. This way, we don’t end up with around 10 new error enum variants, and it’s cleaner.

seemenkina (approved):

PR #362, opened 2025-12-19
fix(ci): add Binaryen installation to CI workflows and Makefile

seemenkina:
Installing software for wasm-opt to avoid timeout errors when CI attempts to install from scratch via wasm-pack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,23 @@ jobs:
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+      - name: Install Binaryen
+        if: matrix.crate == 'rln-wasm'
+        run: |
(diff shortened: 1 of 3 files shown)


vinhtc27:
> Installing software for wasm-opt to avoid timeout errors when CI attempts to install from scratch via wasm-pack

I previously removed Binaryen from the `make installdeps` script due to an inconsistent platform version between macos and linux, but we no longer use wasm-opt from Binaryen.
Instead, current build command use the one from wasm-pack because wasm-opt is now part of wasm-pack. 

I think this failed CI in the master branch is just a cache problem from the GitHub CI infra, maybe remove all the old cache image can resolve this, because every PR tests CI and the Nightly build works perfectly fine. 
So maybe no need to install Binaryen separately in CI.

I rerun the CI in master and it worked: https://github.com/vacp2p/zerokit/actions/runs/20334598808/job/58513904250

PR #373, opened 2026-02-03
Ark groth16 partial

mghazwi:
## Description

The PR includes: 
- Groth16 partial prover which provides a way to generate partial proofs. It uses [ark-groth16](https://github.com/arkworks-rs/groth16) and the implementation is based on [Groth16 prover](https://github.com/logos-storage/nim-groth16) . 
- Integration into zerokit RLN, which includes: evaluation and calculation of the partial witness (known witness that don't change much), and functions to generate partial rln proofs (and finishing these proofs). 

## Testing
- Tests included in the `test/protocol` to both check end-to-end works and check that a finished partial proof, equals to a full proof. 
- Benchmarks: partial proof, finish proof, full proof

### DIFF
--- a/rln/Cargo.toml
+++ b/rln/Cargo.toml
@@ -94,6 +94,7 @@ harness = false
 [[bench]]
 name = "partial_proof"
 harness = false
+required-features = ["pmtree-ft"]
 
 [package.metadata.docs.rs]
 all-features = true


ark-groth16-partial/src/partial_prover.rs
```diff
@@ -0,0 +1,340 @@
...
+    h: &[E::ScalarField],
+    r: E::ScalarField,
+    s: E::ScalarField,
+) -> R1CSResult<Proof<E>> {
+    let num_inputs = pk.vk.gamma_abc_g1.len(); // this includes the "1" input
+    let num_aux = pk.l_query.len();
+    let expected_len = num_inputs + num_aux - 1;
+
+    assert_eq!(full_assignment.len(), expected_len, "assignment length mismatch");
+    assert_eq!(partial.mask.len(), expected_len, "mask length mismatch");
```
seemenkina: could you please return proper error instead of panic in the core library code

ark-groth16-partial/src/partial_prover.rs
```diff
@@ -0,0 +1,340 @@
...
+) -> PartialProof<E> {
+    let num_inputs = pk.vk.gamma_abc_g1.len(); // this includes the "1" input
+    let num_aux = pk.l_query.len();
+    let expected_len = num_inputs + num_aux - 1;
+
+    assert_eq!(
+        partial_assignment.values.len(),
+        expected_len,
+        "partial assignment has wrong length"
+    );
```
seemenkina: same here, please return error

rln/src/protocol/witness.rs
```diff
@@ -28,6 +28,16 @@ pub struct RLNWitnessInput {
     external_nullifier: Fr,
 }
 
+/// Partial witness input for RLN proof precalculation.
+/// Contains the non-changing inputs used to precompute a partial proof:
+#[derive(Debug, PartialEq, Clone)]
+pub struct RLNPartialWitnessInput {
```
seemenkina, line 34: let's implement  `impl From<&RLNWitnessInput> for RLNPartialWitnessInput {}` to avoid duplication code of transfer RLNWitnessInput to RLNPartialWitnessInput

ark-groth16-partial/README.md
```diff
@@ -0,0 +1,7 @@
+# Groth16 Partial Proof Generator
```
seemenkina: can you also please move ark-groth16-partial inside rln/src folder because in other case we can't publish this part in new release as it separate member of rln project right now

seemenkina:
@mghazwi could you please also rebase to master branch and resolve all conflict

rln/src/partial_prover.rs
```diff
@@ -0,0 +1,349 @@
...
+    let mask = partial_assignment
+        .values
+        .iter()
+        .map(|v| v.is_some())
+        .collect::<Vec<_>>();
+
+    let mut a1_points = Vec::new();
+    let mut scalars = Vec::new(); // these are the public input + known witnesses
+    let mut b1_points = Vec::new();
+    let mut b2_points = Vec::new();
```
seemenkina, line 133: can we specify capacity of the vector on this step? There are more initiation in the next code also

seemenkina (commented):
Please run the linter also, because I cannot approve if there is some CI errors

rln/src/partial_prover.rs
mghazwi: the capacity is not known in advance and depends on the number of known witnesses. It does have a max value of size `partial_assignment.values.len()` but then I think we would probably be using more memory than needed.

mghazwi:
@seemenkina @vinhtc27 I have pushed the latest changes to the PR and it should be ready now. 
Please let me know if there are any issues, thanks.

vinhtc27 (approved):
LGTM, Thankyou!

PR #375, opened 2026-02-05
feat: implement multi-message-id feature

vinhtc27:
## Description

- Refactored `RLNWitnessInput` and `RLNProofValues` into struct/enum (`SingleV1` / `MultiV1`) to cleanly support both standard and multi-message-id modes.
- Added version byte retrieval methods for `RLNProof` and `RLNWitnessInput`.
- Enhanced `ProtocolError` with new variants: `EmptyMessageIds`, `DuplicateMessageIds`, `MissingSelectorUsed`, `NoActiveSelectorUsed`, and `FieldLengthMismatch`.
- Read `MAX_OUT` value directly from the graph file instead of hardcoding.
- Added checks for `message_ids`, `selector_used`, `path_elements`, `identity_path_index` in proof gen api.
- Extended FFI (C/Nim/Rust) and WASM bindings with multi-message-id support.
- Added multi-message-id CLI example and updated existing examples.
- Added multi-message-id circuit resources (`max_out_4`) for tree depth 10 and 20; removed tree depth 16 and 24.
- Updated CI with clippy checks and nightly release for the `multi-message-id` feature flag.
- Version bump for dependencies in all modules.
- Exposed compute_id_secret as a public API and updated README on how to use the slashing cross feature.
## Testing

- Added multi-message-id test cases for FFI, protocol, and public API.
- Added compatibility tests to verify new witness/proof types against old serialized data.

## Checklist

- [x] I have run the CI coverage report. Add the `run-coverage` label to this PR to enable it.


rln-wasm/Makefile.toml
```diff
@@ -44,6 +60,45 @@ args = [
...
+    "nightly",
+    "wasm-pack",
+    "build",
+    "--release",
+    "--target",
+    "web",
+    "--scope",
+    "waku",
+    "--features",
+    "parallel,multi_message_id",
```
seemenkina: here wrong feature flag, so it didn't build 

simplified test: 
```
evb:rln-wasm (multi-message-id-burn)$ cargo build --target wasm32-unknown-unknown --features parallel,multi_message_id
error: the package 'rln-wasm' does not contain this feature: multi_message_id
help: there is a similarly named feature: multi-message-id
```

rln/src/ffi/ffi_rln.rs
```diff
@@ -379,6 +484,70 @@ pub fn ffi_rln_proof_values_get_external_nullifier(
...
+                ys.iter()
+                    .map(|fr| CFr::from(*fr))
+                    .collect::<Vec<_>>()
+                    .into(),
+            ),
+            err: None,
+        },
+        None => CResult {
+            ok: None,
+            err: Some("ys field is None, use ffi_rln_proof_values_get_ys instead".into()),
```
seemenkina: maybe here should be `ffi_rln_proof_values_get_y` ?

rln/src/protocol/proof.rs
```diff
@@ -28,50 +33,138 @@ pub struct RLNProof {
...
-    bytes
+    #[cfg(feature = "multi-message-id")]
+    {
+        let capacity = FR_BYTE_SIZE
+            * (5 + rln_proof_values.ys.as_ref().map_or(0, |v| v.len())
+                + rln_proof_values.nullifiers.as_ref().map_or(0, |v| v.len()))
+            + rln_proof_values
+                .selector_used
+                .as_ref()
+                .map_or(0, |v| v.len());
```
seemenkina: Here you missed 8 bytes for each vector, because we construct vectors as  [v.len, v0, v1, ..., vn-1]

rln/src/protocol/proof.rs
```diff
@@ -28,50 +33,138 @@ pub struct RLNProof {
...
+        let mut bytes = Vec::with_capacity(capacity);
```
seemenkina: Here you missed 8 bytes for each vector, because we construct vectors as [v.len, v0, v1, ..., vn-1]

rln/src/protocol/slashing.rs
```diff
@@ -46,10 +46,64 @@ pub fn recover_id_secret(
...
+                    for (j, (nullifier_j, &used_j)) in
+                        nullifiers2.iter().zip(selector_used2.iter()).enumerate()
+                    {
+                        if !used_j {
+                            continue;
+                        }
+                        if nullifier_i == nullifier_j {
+                            let share1 = (rln_proof_values_1.x, ys1[i]);
+                            let share2 = (rln_proof_values_2.x, ys2[j]);
+                            return compute_id_secret(share1, share2);
```
seemenkina: I think here we should add some checks like  ys1.len() >= nullifiers1.len(), because in other case we have OOB panic

rln/src/protocol/proof.rs
```diff
@@ -28,50 +33,138 @@ pub struct RLNProof {
...
+    #[cfg(feature = "multi-message-id")]
+    pub nullifiers: Option<Vec<Fr>>,
     pub root: Fr,
-    // Public Inputs:
+    // Public inputs:
     pub x: Fr,
     pub external_nullifier: Fr,
+    #[cfg(feature = "multi-message-id")]
+    pub selector_used: Option<Vec<bool>>,
 }
```
seemenkina, line 175: So here I have a suggestion to use additional enum instead of repeat field:

```
pub enum RLNOutputs {
      Single { y: Fr, nullifier: Fr },
      #[cfg(feature = "multi-message-id")]
      Multi { ys: Vec<Fr>, nullifiers: Vec<Fr>, selector_used: Vec<bool> },
  }

  pub struct RLNProofValues {
      pub outputs: RLNOutputs,
      pub root: Fr,
      pub x: Fr,
      pub external_nullifier: Fr,
  }
```

In such case we can avoid  all (Some, None) / (None, Some) / _ => Err matches. The _ => arm disappears because the compiler proves all cases are covered. And for example in recover id secret impossible state becomes a compiler error

rln/src/protocol/witness.rs
```diff
@@ -19,38 +24,94 @@ use crate::{
...
+    #[cfg(feature = "multi-message-id")]
+    message_ids: Option<Vec<Fr>>,
     path_elements: Vec<Fr>,
     identity_path_index: Vec<u8>,
+    // Public inputs
     x: Fr,
     external_nullifier: Fr,
+    #[cfg(feature = "multi-message-id")]
+    selector_used: Option<Vec<bool>>,
 }
```
seemenkina, line 56: Same idea as I described in proof.rs: 

```
  pub enum RLNMessageInput {
      Single { message_id: Fr },
     #[cfg(feature = "multi-message-id")]
      Multi { message_ids: Vec<Fr>, selector_used: Vec<bool> },
  }

  pub struct RLNWitnessInput {
      identity_secret: IdSecret,
      user_message_limit: Fr,
      message_input: RLNMessageInput,
      path_elements: Vec<Fr>,
      identity_path_index: Vec<u8>,
      x: Fr,
      external_nullifier: Fr,
  }
```

rln/src/protocol/slashing.rs
```diff
@@ -46,10 +46,64 @@ pub fn recover_id_secret(
...
+        let share1 = (rln_proof_values_1.x, rln_proof_values_1.y);
+        let share2 = (rln_proof_values_2.x, rln_proof_values_2.y);
+        compute_id_secret(share1, share2)
+    }
+
+    #[cfg(feature = "multi-message-id")]
+    {
+        match (
+            (&rln_proof_values_1.y, &rln_proof_values_1.ys),
+            (&rln_proof_values_2.y, &rln_proof_values_2.ys),
```
seemenkina: btw, this match doesn't cover case like (Single, Multi) and (Multi, Single), but maybe this case make sense generally? 

What I mean, a Single proof is mathematically a Multi proof with one active slot. You can normalize both sides before matching, and then slashing has a single code path for all combinations

```
  fn normalize(outputs: &RLNOutputs) -> (Vec<Fr>, Vec<Fr>, Vec<bool>) {
      match outputs {
          RLNOutputs::Single { y, nullifier } => {
              (vec![*y], vec![*nullifier], vec![true])
          }
          #[cfg(feature = "multi-message-id")]
          RLNOutputs::Multi { ys, nullifiers, selector_used } => {
              (ys.clone(), nullifiers.clone(), selector_used.clone())
          }
      }
  }

  pub fn recover_id_secret(
      proof_values_1: &RLNProofValues,
      proof_values_2: &RLNProofValues,
  ) -> Result<IdSecret, ProtocolError> {
      if proof_values_1.external_nullifier != proof_values_2.external_nullifier {
          return Err(ProtocolError::ExternalNullifierMismatch(...));
      }

      let (ys1, nullifiers1, su1) = normalize(&proof_values_1.outputs);
      let (ys2, nullifiers2, su2) = normalize(&proof_values_2.outputs);

      for (i, (nullifier_i, &used_i)) in nullifiers1.iter().zip(su1.iter()).enumerate() {
          if !used_i {
              continue;
          }
          for (j, (nullifier_j, &used_j)) in nullifiers2.iter().zip(su2.iter()).enumerate() {
              if !used_j {
                  continue;
              }
              if nullifier_i == nullifier_j {
                  return compute_id_secret(
                      (proof_values_1.x, ys1[i]),
                      (proof_values_2.x, ys2[j]),
                  );
              }
          }
      }

      Err(ProtocolError::IdSecretRecovery)
  }
```

wdyt, does it make sense for you?

rln/src/protocol/witness.rs
```diff
@@ -19,38 +24,94 @@ use crate::{
...
         // Message ID range check
+        #[cfg(not(feature = "multi-message-id"))]
         if message_id >= user_message_limit {
             return Err(ProtocolError::InvalidMessageId(
                 message_id,
                 user_message_limit,
             ));
         }
 
+        #[cfg(feature = "multi-message-id")]
```
seemenkina, line 109: circuit is compiled with a fixed MAX_OUT=4. The circuit expects exactly 4 messageId signals and 4 selectorUsed signals as inputs. The Rust code has no constant representing MAX_OUT and performs no length validation against it.

rln/src/protocol/witness.rs
```diff
@@ -19,38 +24,94 @@ use crate::{
...
+                    }
+
+                    for (message_id, used) in message_ids.iter().zip(selector_used) {
+                        if *used && *message_id >= user_message_limit {
+                            return Err(ProtocolError::InvalidMessageId(
+                                *message_id,
+                                user_message_limit,
+                            ));
+                        }
+                    }
```
seemenkina: Here, I have a question, do we need some check that all message ids are distinct? 

because looks like now that such situation valid:

  message_ids:   [5, 5, 5, 5]
  selector_used: [1, 1, 1, 1]

and if we need such assumption we also need changes in circuit as well

rln/src/protocol/witness.rs
```diff
@@ -19,38 +24,94 @@ use crate::{
...
+    message_id: Option<Fr>,
+    #[cfg(feature = "multi-message-id")]
+    message_ids: Option<Vec<Fr>>,
     path_elements: Vec<Fr>,
     identity_path_index: Vec<u8>,
+    // Public inputs
     x: Fr,
     external_nullifier: Fr,
+    #[cfg(feature = "multi-message-id")]
+    selector_used: Option<Vec<bool>>,
```
seemenkina: Because the RFC implies at least one used slot (0 < i <= max_out and j = max_out - i), and says NULL outputs “do not represent valid proofs,” but the current Rust/circuit logic never enforces any(selector_used) == true, the implementation currently accepts all-zero-selector proofs (all y/nullifier = 0) that consume no burn, which is a spec mismatch.

rln/src/protocol/witness.rs
vinhtc27: Updated Zerokit to include these checks. I’ll also update the Circom circuit to enforce them.

rln/src/protocol/witness.rs
vinhtc27: Same as above.

rln/src/protocol/witness.rs
vinhtc27: I tried a few different ways to handle this. There are a couple of ways we could enforce the MAX_OUT constant for the multi-message-id feature.

The first option is to use const generics for a type like RLNWitnessInput, so it’s handled at compile time. Then we could check the length internally in the new function. The problem is that with WASM and FFI, we can’t really rely on those compile-time features. The value still has to be passed as a parameter at runtime, which kind of defeats the whole purpose of using const generics in the first place.

The second option is to use a DEFAULT_MAX_OUT value. But that’s obviously very strict and wouldn’t work if a user wants to use a custom circuit with a different MAX_OUT. We could make MAX_OUT a global static modifiable value, but that introduces issues if users create different instances with different circuits. That could lead to atomicity problems.

So I went with a more flexible approach: deriving the MAX_OUT value from the length of the message_id field in the graph.bin file:
https://github.com/vacp2p/zerokit/blob/c52a1d2de00d8ac0dd2a541a189dba139a264dfa/rln/src/circuit/mod.rs#L144

I also exposed an API so users can access to have length check it if needed, like in this example:
https://github.com/vacp2p/zerokit/blob/c52a1d2de00d8ac0dd2a541a189dba139a264dfa/rln-cli/src/examples/multi_message_id.rs#L150

This approach still has a downside: it requires users to make sure they pass the correct length into the new API. But it’s still a valid approach because I check the length before performing witness calculation and proof generation:
https://github.com/vacp2p/zerokit/blob/c52a1d2de00d8ac0dd2a541a189dba139a264dfa/rln/src/public.rs#L587

In the case of rln-wasm, where there’s no graph data available, the witness calculation step will still reject and return an error if the length isn’t correct. I tested this in the Node.js example, and it fails with a clear error message.

WDYT?

rln/src/protocol/proof.rs
seemenkina: I think we should reconsider the enum approach for RLNProofValues and RLNWitnessInput.

The issue is that root, x, external_nullifier are identical in both variants, but because they live inside the enum, every accessor needs a match just to reach them. Same thing in serialization — we're writing the same fields twice across both arms. Since we use #[cfg] gating, only one variant ever exists at compile time, so the enum isn't giving us anything for those shared fields.

If we pull the common fields into a struct and keep the enum only for what actually differs, it cleans up a lot:

```
pub enum RLNOutputs {
    Single { y: Fr, nullifier: Fr },
    #[cfg(feature = "multi-message-id")]
    Multi { ys: Vec<Fr>, nullifiers: Vec<Fr>, selector_used: Vec<bool> },
}

pub struct RLNProofValues {
    pub outputs: RLNOutputs,
    pub root: Fr,
    pub x: Fr,
    pub external_nullifier: Fr,
}
```

Then self.root just works, serialization writes the common prefix once, and slashing doesn't have to destructure the whole thing to get x. Same idea applies to RLNWitnessInput.

Also worth noting: if we ever remove the #[cfg] gates to support both variants at runtime, the full enum approach has a real cost — Rust sizes an enum to its largest variant, so every SingleV1 instance would carry the memory footprint of MultiV1 with its three Vecs. The struct+enum approach avoids that since the shared fields are always just struct fields and only the small RLNOutputs enum pays the largest-variant penalty.

rln/src/public.rs
seemenkina, line 636: you miss max out validation here

rln/src/circuit/mod.rs
```diff
@@ -117,21 +129,39 @@ pub fn graph_from_raw(
...
-    }
+
+        depth
+    };
+
+    #[cfg(all(feature = "multi-message-id", not(target_arch = "wasm32")))]
+    let max_out = input_mapping
+        .get("messageId")
+        .map(|(_, len)| *len)
+        .unwrap_or_default();
```
seemenkina: I'm not sure that default 0  here is a good idea

 If the graph.bin file doesn't contain a messageId signal (e.g., user loads an old single-message circuit with the multi-message-id feature enabled), max_out silently becomes 0. This will cause confusing FieldLengthMismatch errors downstream. I suggest: 
```
let max_out = input_mapping
    .get("messageId")
    .map(|(_, len)| *len)
    .ok_or(GraphReadError::MissingSignal("messageId"))?;
```

rln/src/public.rs
seemenkina, line 617: let's maybe reuse generate_zk_proof function here to avoid duplicated function - it becomes much easier to understand that rln proof include zk proof + something

rln/src/protocol/proof.rs
vinhtc27: Yes, that makes sense. It’s to improve performance and save memory internally, right? And we’re still keeping all those new getters and setters for external use?

rln/src/protocol/witness.rs
seemenkina: I think we should handle `max_out` the same way we handle `tree_depth` — have the user pass it explicitly at construction time and validate it against the graph.

Right now, `max_out` is silently derived from the graph.bin file with `unwrap_or_default()`. If something goes wrong (wrong circuit loaded, missing signal), it defaults to 0 and you get a confusing error later during proof generation. Compare that with `tree_depth`, where the user declares the expected value upfront and `graph_from_raw` immediately fails with `TreeDepthMismatch` if the circuit doesn't match.

The two parameters work the same way conceptually — both are fixed-size array dimensions that the circuit was compiled with. The user always provides exactly `max_out` message_ids in the witness, and uses `selector_used` to control how many are actually active, just like with tree_depth you always provide the full path and pad the rest. So there's no reason to treat them differently.

For a concrete example of why this matters: imagine someone loads a graph.bin for max_out=4 but accidentally pairs it with a zkey compiled for max_out=8. With the current approach, the graph says 4, the witness passes validation, proof generation runs... and then either silently produces an invalid proof or fails deep in the Groth16 math with an opaque error. With the explicit approach, you catch it right at `RLN::new_with_params()` — before any proof work happens.

Something like:

```rust
pub fn new_with_params(
    tree_depth: usize,
    max_out: usize,        // same pattern as tree_depth
    zkey_data: Vec<u8>,
    graph_data: Vec<u8>,
    tree_config: T,
) -> Result<RLN, RLNError>
```

It's just one extra integer — works fine for FFI and WASM, no generics needed. And it keeps the API consistent: both circuit dimensions are validated the same way, at the same time, with clear error messages.

rln/src/protocol/proof.rs
seemenkina: btw, what the reason of having setters for external use? Previously, we only had a new function, and inside that new function we have validation of all incoming parameters. That means we cannot modify proof values or witness after we create it.

Right now, if you allow modification, it can be situated that we create witness and right after this we modify all values as we need. We lost security in such a case.

My suggestion is to return only for the new function and remove all modifiers, because it doesn't make sense at all. Or, if we really need them, we should copy all validation checks. We still do not solve the issue that we pass some new values. They can pass as a validation check, but they will be wrong.

so what the idea behind setters, only because we don't have access to the enum field?

vinhtc27: It's was for testing only and i was thinking it might helpful if i expose them to the public/ffi/wasm too.
I'll remove them then.

rln/src/public.rs
```diff
@@ -587,8 +617,8 @@ impl RLN {
         &self,
         witness: &RLNWitnessInput,
     ) -> Result<(Proof, RLNProofValues), RLNError> {
-        let proof_values = proof_values_from_witness(witness)?;
         let proof = generate_zk_proof(&self.zkey, witness, &self.graph)?;
+        let proof_values = proof_values_from_witness(witness)?;
```
seemenkina, line 621: I think getting error from proof_values_from_witness is better before possible zk generation

Issue #379, opened 2026-02-18
Unhandled vector capacity overflow - ffi_atomic_operation


romanzac:
The `test_rln_out_of_bounds_ffi` fails at the case `Test atomic operation with valid leaf_index but OOB in indices_to_delete`. This chain of calls will panic: 

` RLN::atomic_operation → PmTree::override_range(0, [1 leaf], [1048586]) → PmTree::remove_indices_and_set_leaves(start=0, leaves=[1 Fr], indices=[1048586]) in [rln/src/pm_tree_adapter.rs](http://localhost:12345/rln/src/pm_tree_adapter.rs:437)` 

The test expects !result.ok (error return), but the panic aborts before returning.

### Impact
Low occurrence, medium impact.

### To reproduce
1. Checkout https://github.com/vacp2p/zerokit/pull/376/commits/c2481af312da0d39f1b82da3494a2aebce1194de
2. cargo make test

### Expected behavior
Return an early error when index is out of bounds. Perhaps `validate all indices <= capacity - 1` and `min(indices) <= start + leaves.len() - 1` in `override_range` or `atomic_operation` before processing.

### Screenshots/logs

```
thread 'test::test_rln_out_of_bounds_ffi' (4113534) panicked at library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
stack backtrace:
   0:        0x102c21598 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h2808c764b780ef77
   1:        0x102c2ee58 - core::fmt::write::h164e2e78980a3344
   2:        0x102bfb878 - std::io::Write::write_fmt::hc5b2a4d160c11c86
   3:        0x102c03e84 - std::sys::backtrace::BacktraceLock::print::h765d16e49e83c7b3
   4:        0x102c06d28 - std::panicking::default_hook::{{closure}}::hf5d7ef6cefa10586
   5:        0x102c06bdc - std::panicking::default_hook::h8781bd6c5c54e4cc
   6:        0x102c07468 - std::panicking::panic_with_hook::h93c775fc227522dd
   7:        0x102c07128 - std::panicking::panic_handler::{{closure}}::h8561e58c9391724a
   8:        0x102c03f90 - std::sys::backtrace::__rust_end_short_backtrace::h55629ed76aaa9d26
   9:        0x102bf44c0 - __rustc[d9b87f19e823c0ef]::rust_begin_unwind
[...]

PR #380, opened 2026-02-18
fix(rln): prevent atomic_operation panics with shared override_range validation

seemenkina:
## Description

This PR fixes panic paths in `override_range`/`atomic_operation` flows by centralizing index/range validation and converting invalid inputs into explicit errors.

Related to: #379 

## Changes

- Added shared validator: `validate_override_range_inputs` in `utils/src/merkle_tree/override_range_validation.rs`.
- Added `EmptyIndicesPolicy` enum (`Allow` / `Reject`) to make adapter-specific empty-index behavior explicit.
- Wired shared validation into:
  - `rln/src/pm_tree_adapter.rs`
  - `utils/src/merkle_tree/full_merkle_tree.rs`
  - `utils/src/merkle_tree/optimal_merkle_tree.rs`
- Replaced unchecked `start + leaves_len` arithmetic with `checked_add` in `set_range` paths for full/optimal trees.

## Testing

- Added regression coverage in `rln/tests/ffi.rs` for:
  - valid `start` with problematic delete-index layout (previously could panic),
  - overflowing `start` (`usize::MAX`) in atomic operation.
- Added unit tests for validator behavior (overflow, incompatible mixed offsets, empty-index policy).

Covered scenarios:
- OOB delete indices now return error (no panic).
- Overflow in `start + leaves_len` now returns error (no panic).
- Empty-index handling is explicit per implementation via `EmptyIndicesPolicy`.

---

## PR Lifecycle

> [!IMPORTANT]
> **Draft PRs** signal that work is still in progress and **will not trigger CI**.
> Only mark your PR as **Ready for review** when you believe it is complete.
> All CI checks **must pass** before requesting a review.

## Code Guidelines

Please keep the following in mind (see [CONTRIBUTING.md](../CONTRIBUTING.md) for full details):

### Commits

- Follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (`feat(rln):`, `fix(utils):`, `chore:`, etc.)
- Use the appropriate scope: `rln`, `rln-cli`, `rln-wasm`, `utils`, `ci`
- GPG-sign your commits

### Error Handling

- **No panics in library code.** Do not use `unwrap()`, `expect()`, or `panic!()`
  in production paths inside `rln/src/` or `utils/src/`.
  The only acceptable exception is an internal invariant that is statically guaranteed — and even then,
  prefer returning an error.
- Use the project's `thiserror`-based error types (`RLNError`, `ProtocolError`, `UtilsError`, etc.)
  and propagate errors with `?`.
- Provide context in error variants (e.g., `InsufficientData { expected, actual }`).
- `unwrap()` is fine in **tests**.

### Code Style

- Run `cargo fmt --all -- --check` to verify formatting (CI enforces this on stable).
- Group imports: std first, then external crates, then local modules (see `rustfmt.toml`).
- Use `pub(crate)` for items that should not be part of the public API.
- Apply `Zeroize` / `ZeroizeOnDrop` to any struct holding secret material.

### Linting (mirrors CI)

CI runs clippy across multiple crate/feature combinations. Run the relevant checks locally before pushing:

```bash
# Default features — workspace root (rln + utils)
cargo clippy --all-targets --tests --release -- -D warnings

# Stateless feature — from rln/
cd rln && cargo clippy --all-targets --tests --release \
  --features=stateless --no-default-features -- -D warnings

# WASM target — from rln-wasm/
cd rln-wasm && cargo clippy --target wasm32-unknown-unknown \
  --tests --release -- -D warnings
```

At minimum, run the default-features check. If your changes touch `stateless` or `rln-wasm`, run those checks as well.

## Checklist

- [x] My PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
- [x] I have linked the related issue(s)
- [x] `cargo fmt --all -- --check` produces no changes
- [x] Clippy passes for all affected crate/feature combinations (see [Linting](#linting-mirrors-ci) above)
- [x] `make test` passes locally
- [x] No new `unwrap()` / `expect()` / `panic!()` in library code
- [x] New code includes appropriate tests (unit / integration / WASM where applicable)
- [x] I have run the CI coverage report — add the `run-coverage` label to enable it
- [x] All CI checks pass and the PR is marked **Ready for review**

### DIFF
--- a/utils/src/merkle_tree/override_range_validation.rs
+++ b/utils/src/merkle_tree/override_range_validation.rs
@@ -70,6 +70,22 @@ mod tests {
     use crate::error::ZerokitMerkleTreeError;
 
     #[test]
+    fn test_validate_override_range_inputs_accepts_valid_inputs() {
+        // indices [0,1] are before start=2, leaves cover [2..=4] (indices 2,3,4), all within capacity
+        let validated = validate_override_range_inputs(
+            2,
+            3,
+            vec![1, 0],
+            1usize << 20,
+            EmptyIndicesPolicy::Allow,
+        )
+        .unwrap();
+        assert_eq!(validated.min_index, Some(0));
+        assert_eq!(validated.max_index, Some(5)); // start + leaves_len
+        assert_eq!(validated.indices, vec![0, 1]); // sorted
+    }
+
+    #[test]
     fn test_validate_override_range_inputs_rejects_start_add_overflow() {
         let err = validate_override_range_inputs(
             usize::MAX,
--- a/utils/benches/merkle_tree_benchmark.rs
+++ b/utils/benches/merkle_tree_benchmark.rs
@@ -1,12 +1,13 @@
 use std::{fmt::Display, str::FromStr, sync::LazyLock};
 
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use tiny_keccak::{Hasher as _, Keccak};
 use zerokit_utils::{
     error::HashError,
+    merkle_tree::validate_override_range_inputs,
     merkle_tree::{
-        FullMerkleConfig, FullMerkleTree, Hasher, OptimalMerkleConfig, OptimalMerkleTree,
-        ZerokitMerkleTree,
+        EmptyIndicesPolicy, FullMerkleConfig, FullMerkleTree, Hasher, OptimalMerkleConfig,
+        OptimalMerkleTree, ZerokitMerkleTree,
     },
 };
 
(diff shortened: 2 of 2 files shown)


utils/src/merkle_tree/override_range_validation.rs
```diff
@@ -0,0 +1,100 @@
...
+        max_index,
+        min_index,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{validate_override_range_inputs, EmptyIndicesPolicy};
+    use crate::error::ZerokitMerkleTreeError;
+
```
sydhds, line 71: I'm wondering if the first test here should not be a valid one? Otherwise we could end with an implementation always returning an error and never returning Ok(...) ?

utils/src/merkle_tree/override_range_validation.rs
```diff
@@ -0,0 +1,100 @@
...
+    if matches!(empty_indices_policy, EmptyIndicesPolicy::Reject) && indices.is_empty() {
+        return Err(ZerokitMerkleTreeError::InvalidIndices);
+    }
+
+    if indices.iter().any(|&i| i >= capacity) {
+        return Err(ZerokitMerkleTreeError::InvalidIndices);
+    }
+
+    indices.sort_unstable();
+    indices.dedup();
```
sydhds, line 36: I'm wondering if we should provide 2 APIS here? One that use this validate_override_range_inputs function and one (~ raw api) that let the user in control (for faster code, assuming no duplicate and sorted indices)?

seemenkina: so you suggested to have validate_override_range_inputs only with checking that indices not bigger than capacity?

utils/src/merkle_tree/override_range_validation.rs
seemenkina: make sense, thank you, I'll add a test

utils/src/merkle_tree/override_range_validation.rs
seemenkina: I thought about this, and it seems to me that we will only complicate development if there are two interfaces and we leave it up to the user to decide which one to choose. 

I also measured how long it takes to validate in the override function on a large number of elements up to 65k, and it turned out to be no more than 1.3% of all calculations. 

I left all numbers in this [commit](https://github.com/vacp2p/zerokit/pull/380/changes/a870061f14309443f0acfdca6872dea2bf9d71a9)

sydhds: Ok so no need to provide a raw api then :)

PR #403, opened 2026-04-14
refactor(PR 0): add message mode and remove multi message id flag

vinhtc27:
## Changes:
- Removed multi-message-id feature flag
- Removed all compiled time gated config flag #cfg
- Update all related examples, documentations, and CI workflows


rln/src/protocol/mode.rs
```diff
@@ -0,0 +1,180 @@
...
+///                  | path_elements<var> | identity_path_index<var> ]
+/// MultiV1:  [ 0x01 | identity_secret<32> | user_message_limit<32>
+///                  | path_elements<var> | identity_path_index<var> ]
+/// ```
+///
+/// **`RLNProofValues`**
+/// ```text
+/// SingleV1: [ 0x00 | root<32> | external_nullifier<32> | x<32> | y<32> | nullifier<32> ]
+/// MultiV1:  [ 0x01 | root<32> | external_nullifier<32> | x<32> | ys<var> | nullifiers<var>
+///                  | selector_used<var> ]
```
seemenkina: the comment actually doesn't satisfy code - 

rln/src/protocol/proof.rs:920-925 — RLNProofValuesSingle field order is root, x, external_nullifier, y, nullifier

so, I think better to have order of the element how it was and how it defined in the rfc. Either update comment or code

rln/src/protocol/mode.rs
```diff
@@ -0,0 +1,180 @@
...
+/// |------------|--------|------|
+/// | `SingleV1` | `0x00` | <https://lip.logos.co/ift-ts/raw/rln-v2> |
+/// | `MultiV1`  | `0x01` | <https://lip.logos.co/ift-ts/raw/multi-message_id-burn-rln> |
+///
+/// ### Wire layouts per type
+///
+/// **`RLNWitnessInput`**
+/// ```text
+/// SingleV1: [ 0x00 | identity_secret<32> | user_message_limit<32> | message_id<32>
+///                  | path_elements<var> | identity_path_index<var> | x<32> | external_nullifier<32> ]
```
seemenkina, line 29: the comment doesn't satisfy the code 

rln/src/protocol/witness.rs:945-953 — RLNWitnessInputSingle field order ends with message_id

so, I think better to have order of the element how it was and how it defined in the rfc. Either update comment or code

rln/src/protocol/proof.rs
seemenkina, line 857: so here general notes again around the order - the circuit's public-input layout is [y, root, nullifier, x, external_nullifier] — fixed by the circuit. When you use Canonical serialisation it happens same to the field ordering in the structure/vector. So check carefully that order is satisfy to the circuit + let's also try to satisfy rfc

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+impl CanonicalDeserializeBE for Vec<bool> {
+    type Error = UtilsError;
+
+    fn deserialize<R: Read>(mut reader: R) -> Result<Self, Self::Error> {
+        let mut len_buf = [0u8; VEC_LEN_BYTE_SIZE];
+        reader.read_exact(&mut len_buf)?;
+        let count = usize::try_from(u64::from_be_bytes(len_buf))?;
+        let mut raw = vec![0u8; count];
+        reader.read_exact(&mut raw)?;
+        Ok(raw.into_iter().map(|b| b != 0).collect())
```
seemenkina: here a byte of value 5 deserializes to true, same as 1. So [0x00, 0x05] and [0x00, 0x01] produce the same vector. I'm not sure if it can produce any issues, but better to have standard way for the bool: 

```
match b {
    0 => Ok(false),
    1 => Ok(true),
    _ => Err(SerializationError::InvalidData),
}
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
+use std::io::{Read, Write};
```
seemenkina, line 7: Let's add comment for the file: 

```
//! Big-endian serialization is implemented manually via [`CanonicalSerializeBE`] /
//! [`CanonicalDeserializeBE`] traits defined here. Little-endian serialization
//! relies on arkworks' [`CanonicalSerialize`] derive for structs; we hand-write
//! it here only for V3 enums (witness, proof values), since arkworks does not
//! derive enum impls.
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    fn serialized_size(&self) -> usize {
+        FR_BYTE_SIZE // identity_secret
+            + FR_BYTE_SIZE // user_message_limit
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.path_elements.len() // path_elements
+            + VEC_LEN_BYTE_SIZE + self.identity_path_index.len() // identity_path_index
+            + FR_BYTE_SIZE // x
+            + FR_BYTE_SIZE // external_nullifier
+            + FR_BYTE_SIZE // message_id
+    }
```
seemenkina, line 366:
```suggestion
    fn serialized_size(&self) -> usize {
        self.identity_secret.serialized_size()
            + self.user_message_limit.serialized_size()
            + self.path_elements.serialized_size()
            + self.identity_path_index.serialized_size()
            + self.x.serialized_size()
            + self.external_nullifier.serialized_size()
            + self.message_id.serialized_size()
    }
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+    fn serialized_size(&self) -> usize {
+        FR_BYTE_SIZE // identity_secret
+            + FR_BYTE_SIZE // user_message_limit
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.path_elements.len() // path_elements
+            + VEC_LEN_BYTE_SIZE + self.identity_path_index.len() // identity_path_index
+            + FR_BYTE_SIZE // x
+            + FR_BYTE_SIZE // external_nullifier
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.message_ids.len() // message_ids
+            + VEC_LEN_BYTE_SIZE + self.selector_used.len() // selector_used
+    }
```
seemenkina, line 416:
```suggestion
    fn serialized_size(&self) -> usize {
        self.identity_secret.serialized_size()
            + self.user_message_limit.serialized_size()
            + self.path_elements.serialized_size()
            + self.identity_path_index.serialized_size()
            + self.x.serialized_size()
            + self.external_nullifier.serialized_size()
            + self.message_ids.serialized_size()
            + self.selctor_used.serialized_size()
    }
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+        self.identity_path_index.serialize(&mut writer)?;
+        Ok(())
+    }
+
+    fn serialized_size(&self) -> usize {
+        FR_BYTE_SIZE // identity_secret
+            + FR_BYTE_SIZE // user_message_limit
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.path_elements.len() // path_elements
+            + VEC_LEN_BYTE_SIZE + self.identity_path_index.len() // identity_path_index
+    }
```
seemenkina, line 460:
```suggestion
    fn serialized_size(&self) -> usize {
        self.identity_secret.serialized_size()
            + self.user_message_limit.serialized_size()
            + self.path_elements.serialized_size()
            + self.identity_path_index.serialized_size()
    }
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+        Ok(())
+    }
+
+    fn serialized_size(&self) -> usize {
+        FR_BYTE_SIZE // root
+            + FR_BYTE_SIZE // x
+            + FR_BYTE_SIZE // external_nullifier
+            + FR_BYTE_SIZE // y
+            + FR_BYTE_SIZE // nullifier
+    }
```
seemenkina, line 602:
```suggestion
    fn serialized_size(&self) -> usize {
        self.x.serialized_size()
            + self.external_nullifier.serialized_size()
            + self.y.serialized_size()
            + self.nullifier.serialized_size()
    }
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    fn serialized_size(&self) -> usize {
+        FR_BYTE_SIZE // root
+            + FR_BYTE_SIZE // x
+            + FR_BYTE_SIZE // external_nullifier
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.ys.len() // ys
+            + VEC_LEN_BYTE_SIZE + FR_BYTE_SIZE * self.nullifiers.len() // nullifiers
+            + VEC_LEN_BYTE_SIZE + self.selector_used.len() // selector_used
+    }
```
seemenkina, line 644:
```suggestion
    fn serialized_size(&self) -> usize {
        self.x.serialized_size()
            + self.external_nullifier.serialized_size()
            + self.ys.serialized_size()
            + self.nullifiers.serialized_size()
            + self.selector_used.serialized_size()
    }
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
+        let bigint = self.into_bigint();
+        let mut buf = [0u8; FR_BYTE_SIZE];
+        for (i, &limb) in bigint.0.iter().rev().enumerate() {
+            buf[i * FR_LIMB_BYTE_SIZE..(i + 1) * FR_LIMB_BYTE_SIZE]
+                .copy_from_slice(&limb.to_be_bytes());
+        }
+        writer.write_all(&buf).map_err(UtilsError::IoError)
+    }
```
seemenkina, line 76: The new BE implementation looks good. Just want to flag: while we still keep the old fr_to_bytes_be in utils.rs, we have two different implementations of the same encoding. Old paths like WASM and FFI may still use the old one. If arkworks changes something internally, new code adapts but the old one may not — and then they produce different bytes for the same Fr. We should be careful around this until the old
serialisation is fully removed and not reuse it in new  WASM, FFI

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+}
+
+impl CanonicalDeserializeBE for IdSecret {
+    type Error = UtilsError;
+
+    fn deserialize<R: Read>(mut reader: R) -> Result<Self, Self::Error> {
+        let mut buf = Zeroizing::new([0u8; FR_BYTE_SIZE]);
+        reader.read_exact(buf.as_mut())?;
+        let mut limbs = [0u64; FR_BYTE_SIZE / FR_LIMB_BYTE_SIZE];
```
seemenkina:
```suggestion
        let mut limbs = Zeroizing::new([0u64; FR_BYTE_SIZE / FR_LIMB_BYTE_SIZE]);
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+        }
+        Ok(Fr::from(bigint))
+    }
+}
+
+impl CanonicalSerializeBE for IdSecret {
+    type Error = UtilsError;
+
+    fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
+        let mut bigint = self.into_bigint();
```
seemenkina: arkworks actually provide zeroizing - I wrote some tests and the work with next definition; so no need to do it manually

```suggestion
       let bigint = Zeroizing::new(fr.into_bigint());
```

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,656 @@
...
+    fn deserialize<R: Read>(mut reader: R) -> Result<Self, Self::Error> {
+        let mut buf = Zeroizing::new([0u8; FR_BYTE_SIZE]);
+        reader.read_exact(buf.as_mut())?;
+        let mut limbs = [0u64; FR_BYTE_SIZE / FR_LIMB_BYTE_SIZE];
+        for (i, limb) in limbs.iter_mut().enumerate() {
+            let start = i * FR_LIMB_BYTE_SIZE;
+            *limb = u64::from_be_bytes(buf[start..start + FR_LIMB_BYTE_SIZE].try_into()?);
+        }
+        limbs.reverse();
+        let bigint = ark_ff::BigInt(limbs);
```
seemenkina:
```suggestion
       let bigint = Zeroizing::new(fr.into_bigint());
```

rln/src/protocol/witness.rs
```diff
@@ -958,3 +934,103 @@ pub(super) fn inputs_for_partial_witness_calculation(
...
+    pub(crate) user_message_limit: Fr,
+    pub(crate) path_elements: Vec<Fr>,
+    pub(crate) identity_path_index: Vec<u8>,
+    pub(crate) x: Fr,
+    pub(crate) external_nullifier: Fr,
+    pub(crate) message_id: Fr,
+}
+
+impl RLNWitnessInputSingle {
+    pub fn new(
```
seemenkina, line 954: don't forget to add validation of inputs into new functions

rln/src/protocol/witness.rs
```diff
@@ -279,14 +294,9 @@ impl RLNPartialWitnessInput {
...
-        #[cfg(not(feature = "multi-message-id"))]
-        {
-            SerializationVersion::SingleV1.into()
-        }
-        #[cfg(feature = "multi-message-id")]
-        {
-            SerializationVersion::MultiV1.into()
-        }
+        // TODO: new enum for partial witness instead of reusing SingleV1 version byte, which is technically not correct
+        // TODO: current master branch return SingleV1 or MultiV1 version byte based compile-time feature flag
```
seemenkina: I think this todo doesn't make sense in V3, right?

rln/src/protocol/witness.rs
seemenkina, line 301: you didn't add same impl for the new types, I think it good to have it as well

rln/src/public.rs
```diff
@@ -762,3 +763,88 @@ impl RLN {
...
+    pub fn insert_leaf(&mut self, _index: usize, _leaf: Fr) -> Result<(), RLNError> {
+        todo!()
+    }
+}
+
+impl<Tree, ZkProof: RLNZkProof> RLNV3<Tree, ZkProof> {
+    pub fn generate_proof(
+        &self,
+        _witness: ZkProof::Witness,
+    ) -> Result<(ZkProof::Proof, ZkProof::Values), RLNError> {
```
seemenkina: The trait RLNZkProof says Self::Error is an associated type, but the wrapper hardcodes RLNError, so we need to do like this  

`pub fn generate_proof(...) -> Result<..., ZkProof::Error>`

seemenkina: and the same for another function

rln/src/public.rs
```diff
@@ -762,3 +763,88 @@ impl RLN {
         Ok(true)
     }
 }
+
+pub struct RLNV3<Mode, ZkProof> {
```
seemenkina, line 767: So, I suppose here just part of the function that would be implemented, right?

rln/src/protocol/mode.rs
vinhtc27: Same as above

rln/src/protocol/serialize.rs
vinhtc27: Yes this will be remove completely in the future, i plan to remove this utils file since we gonna no longer need it

rln/src/protocol/witness.rs
vinhtc27: I added it in the PR4, merge later to avoid conflict.

rln/src/protocol/witness.rs
vinhtc27: Yes u right, the whole serialize format gonna change anyway

rln/src/public.rs
vinhtc27: Yes this will be in PR4, its just all dummy methods for now, didnt cover all old public API yet

seemenkina (approved):

PR #404, opened 2026-04-14
refactor(PR 1): all new RLN types, traits, generics for new proposed architecture

vinhtc27:


rln/src/public.rs
```diff
@@ -764,3 +763,88 @@ impl RLN {
         Ok(true)
     }
 }
+
+pub struct RLNV3<State, ZKBackend> {
```
sydhds: I'd rather go with "Mode" here. I think it is more explicit than "State" here.

rln/src/public.rs
```diff
@@ -764,3 +763,88 @@ impl RLN {
...
+    pub(crate) _zk: ZKBackend,
```
sydhds: Also not sure about the naming of "ZKBackend". I used "ZKP" (for ZkProof) in the POC. I think it makes more sense here and is shorted no?

rln/src/public.rs
```diff
@@ -764,3 +763,88 @@ impl RLN {
...
+    pub(crate) state: State,
+}
+
+impl<ZKBackend> RLNV3<Stateless, ZKBackend> {
+    pub fn new_stateless(zk: ZKBackend) -> Self {
```
sydhds: Maybe we could use new here instead of new_stateless and new_stateful. The dev already knows about the "State" (or "Mode") here

rln/src/protocol/state.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub fn tree_mut(&mut self) -> &mut T {
+        &mut self.tree
+    }
+
+    pub fn into_tree(self) -> T {
+        self.tree
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
sydhds: Do we need so much derive here like Copy? PartialEq? Eq?

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,25 @@
+use std::io::{Read, Write};
+
+pub trait CanonicalSerializeBE {
+    type Error;
+
+    fn serialize_be<W: Write>(&self, writer: W) -> Result<(), Self::Error>;
+
+    fn serialized_size_be(&self) -> usize;
```
sydhds: no need to add "_be" prefix here as well

rln/src/protocol/zk.rs
```diff
@@ -0,0 +1,100 @@
...
+};
+
+pub trait RLNZkProof {
+    type Witness: CanonicalSerialize
+        + CanonicalDeserialize
+        + CanonicalSerializeBE
+        + CanonicalDeserializeBE;
+    type Proof: CanonicalSerialize
+        + CanonicalDeserialize
+        + CanonicalSerializeBE
```
sydhds, line 18: I don't think we could provided BE ser for Proof (or at least not in a efficient way). Idea was to provide only LE ser here

rln/src/protocol/zk.rs
```diff
@@ -0,0 +1,100 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+use crate::{
+    circuit::{ArkGroth16Backend, PartialProof, Proof},
+    error::RLNError,
+    prelude::RLNPartialWitnessInputV3,
+    protocol::{RLNProofValuesV3, RLNWitnessInputV3},
+};
+
+pub trait RLNZkProof {
+    type Witness: CanonicalSerialize
```
sydhds: All trait bounds should be WithFlags (for instance: CanonicalSerializeWithFlags) as we discussed in Discord no?

rln/src/protocol/zk.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+pub trait RecoverSecret<Rhs = Self> {
+    type Error;
+
+    fn recover_secret(&self, other: &Rhs) -> Result<Fr, Self::Error>;
+}
+
+pub trait RLNPartialZkProof: RLNZkProof {
+    type PartialWitness;
```
sydhds: Should have trait bounds for SER/DER no?

rln/src/protocol/zk.rs
```diff
@@ -0,0 +1,100 @@
...
+        + CanonicalDeserialize
+        + CanonicalSerializeBE
+        + CanonicalDeserializeBE;
+    type Values: RecoverSecret
+        + TryFrom<Self::Witness>
+        + CanonicalSerialize
+        + CanonicalDeserialize
+        + CanonicalSerializeBE
+        + CanonicalDeserializeBE;
+    type Error;
```
sydhds, line 30: Not sure if we should provide the same error for generate_proof & verify here. Wdyt?

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,25 @@
+use std::io::{Read, Write};
+
+pub trait CanonicalSerializeBE {
+    type Error;
+
+    fn serialize_be<W: Write>(&self, writer: W) -> Result<(), Self::Error>;
```
sydhds: No need to add "_be" prefix here

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    fn serialized_size_be(&self) -> usize;
+
+    fn to_bytes_be(&self) -> Result<Vec<u8>, Self::Error> {
```
sydhds: Not sure why we need this as we already have serialize func no?

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,25 @@
...
+        Ok(buf)
+    }
+}
+
+pub trait CanonicalDeserializeBE: Sized {
+    type Error;
+
+    fn deserialize_be<R: Read>(reader: R) -> Result<Self, Self::Error>;
+
+    fn from_bytes_be(bytes: &[u8]) -> Result<Self, Self::Error> {
```
sydhds: Same comment as previous

rln/src/protocol/proof.rs
```diff
@@ -893,3 +900,182 @@ pub fn verify_zk_proof(
...
+        _reader: R,
+        _compress: ark_serialize::Compress,
+        _validate: ark_serialize::Validate,
+    ) -> Result<Self, ark_serialize::SerializationError> {
+        todo!()
+    }
+}
+
+impl TryFrom<RLNWitnessInputV3> for RLNProofValuesV3 {
+    type Error = ProtocolError;
```
sydhds: I things ProtocolError is not the right error here. I'd rather define a common error that covers all of these TryFrom impl

rln/src/protocol/proof.rs
```diff
@@ -893,3 +900,182 @@ pub fn verify_zk_proof(
...
+
+    fn try_from(_witness: RLNWitnessInputV3) -> Result<Self, Self::Error> {
+        todo!()
+    }
+}
+
+impl RecoverSecret for RLNProofValuesV3 {
+    type Error = ProtocolError;
```
sydhds: Same here. A common error for all RecoverSecret impl

rln/src/protocol/mode.rs
```diff
@@ -152,3 +152,7 @@ impl From<&Graph> for MessageMode {
         }
     }
 }
+
+pub struct SingleMessage;
```
sydhds: Not sure why these structs are here for?

sydhds (changes requested):

rln/src/protocol/serialize.rs
vinhtc27: We need this suffix to be distinct from the serialize function from arkworks

rln/src/protocol/zk.rs
vinhtc27: Ill figure out the detail for the wrapper error in next PR, lets keep it simple here

rln/src/public.rs
vinhtc27: Let’s decide on the naming after these PRs are mostly done.
We’ll still need to rename the struct anyway.

rln/src/protocol/proof.rs
vinhtc27: Do you mean using a concrete Error type for CanonicalSerializeBE instead of allowing a custom Error type, like above?

rln/src/protocol/proof.rs
vinhtc27: Same question as above

rln/src/protocol/zk.rs
vinhtc27: Lets just leave it here, I'll remove it later in PR3 if we can't figure this out for big endian.

Also, I couldn’t implement `CanonicalSerializeWithFlags` (and `CanonicalDeserializeWithFlags`) for `Proof`. I got the error: `only traits defined in the current crate can be implemented for types defined outside of the crate`
However, `PartialProof` still works. I’ve added a TODO to investigate this further later.

rln/src/public.rs
sydhds: Nothing has been changed on this one no?

rln/src/public.rs
sydhds: I would rather have the correct naming first as it will make the next PR easy to read

rln/src/protocol/zk.rs
sydhds: This is also really important and should be fix on this PR. I think a new type pattern should solve your issue here ?

rln/src/protocol/zk.rs
sydhds: This PR is about having the correct struct, types and traits so I guess it's the right time here...

rln/src/protocol/serialize.rs
sydhds: Still thinks this is weird to have the _be prefix, please see this example: https://doc.rust-lang.org/rust-by-example/trait/disambiguating.html

rln/src/protocol/proof.rs
sydhds: I would keep the type Error = MyError here and define MyError as specific to these TryFrom impl

rln/src/public.rs
vinhtc27: I think `ZkProof` will be more consistent with `RLNZkProof` and `RLNPartialZkProof`

rln/src/protocol/zk.rs
vinhtc27: I created the `ProofV3` type for this

rln/src/protocol/zk.rs
vinhtc27: We currently have a separate `VerifyError` type, but it still lives under `RLNError`, so we don’t need to make it too generic, it’s somewhat redundant complexity and annoying for callers.

rln/src/protocol/proof.rs
vinhtc27: There are many places in the new refactor skeleton that have the same pattern:
```rust
Error = ProtocolError
```
I think we’re going to have a PR to extend Error types anyway, so let’s just keep this as a placeholder for now.

rln/src/protocol/serialize.rs
```diff
@@ -0,0 +1,15 @@
+use std::io::{Read, Write};
+
+pub trait CanonicalSerializeBE {
```
sydhds, line 5: This should match the CanonicalSerializeWithFlags trait (user should be able to pass a flag)

rln/src/circuit/mod.rs
```diff
@@ -104,9 +112,145 @@ pub type G2Projective = ArkG2Projective;
 /// Groth16 proof for the BN254 curve.
 pub type Proof = ArkProof<Curve>;
 
+/// Groth16 proof with additional flags for the BN254 curve.
+pub type ProofV3 = ProofWithFlags;
+
+type RawProof = ArkProof<Curve>;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct ProofWithFlags(RawProof);
```
sydhds: Cannot we go directly with: `struct ZkProof(ArkProof<Curve>)` here. Not sure, that RawProof name is meaningful here. Also ProofWithFlags name makes little sense here I think (WithFlags is for Serialization only and does not relate to the Proof) wdyt?

sydhds (changes requested):

sydhds (approved):

PR #407, opened 2026-04-17
refactor(PR 2): implement LE serialization for all new types

vinhtc27:


rln/src/circuit/mod.rs
```diff
@@ -401,8 +425,53 @@ impl ArkGroth16Backend {
...
+    use ark_serialize::{
+        CanonicalDeserialize, CanonicalDeserializeWithFlags, CanonicalSerialize,
+        CanonicalSerializeWithFlags, EmptyFlags,
+    };
+
     use super::*;
 
+    fn make_proof_v3() -> ProofV3 {
+        ProofWithFlags(ArkProof {
+            a: G1Affine::default(),
```
sydhds: Can ArkProof::Default be used here?

rln/src/circuit/mod.rs
```diff
@@ -401,8 +425,53 @@ impl ArkGroth16Backend {
...
+            b: G2Affine::default(),
+            c: G1Affine::default(),
+        })
+    }
+
+    fn make_partial_proof() -> PartialProof {
+        PartialProof {
+            mask: vec![true, false],
+            partial_pi_a: G1Projective::default(),
```
sydhds: Can use ..Default::default() here no?

rln/src/circuit/mod.rs
```diff
@@ -401,8 +425,53 @@ impl ArkGroth16Backend {
...
+            partial_rho: G1Projective::default(),
+            partial_pi_b: G2Projective::default(),
+            partial_pi_c: G1Projective::default(),
+        }
+    }
+
+    #[test]
+    fn test_proof_v3_le_compressed_roundtrip() {
```
sydhds: Here you are writing unit tests for serialized_compressed and serialized_compressed_with_flags. But if seriazlized_compressed_with_flags use serialized_compressed, there is no need to have both unit tests here I think

rln/src/circuit/mod.rs
```diff
@@ -434,4 +503,52 @@ mod test {
         .unwrap();
         assert!(matches!(err, GraphReadError::MaxOutMismatch { .. }));
     }
+
+    #[test]
+    fn test_proof_v3_le_with_flags_roundtrip() {
```
sydhds: Test is only done with EmpyFlags here? Maybe unit testing with a "real" flags would be helpful?

rln/src/protocol/proof.rs
```diff
@@ -928,53 +928,89 @@ impl RecoverSecret for RLNProofValuesV3 {
...
+                1u8.serialize_with_mode(&mut writer, compress)?;
+                inner.serialize_with_mode(&mut writer, compress)
+            }
+        }
     }
 
-    fn serialized_size(&self, _compress: Compress) -> usize {
-        todo!()
+    fn serialized_size(&self, compress: Compress) -> usize {
+        1 + match self {
```
sydhds: Would be clearer with u8 byte size I think or a constant

rln/src/protocol/proof.rs
```diff
@@ -928,53 +928,89 @@ impl RecoverSecret for RLNProofValuesV3 {
...
         &self,
-        _writer: W,
-        _compress: Compress,
+        mut writer: W,
+        compress: Compress,
     ) -> Result<(), SerializationError> {
-        todo!()
+        match self {
+            RLNProofValuesV3::Single(inner) => {
+                0u8.serialize_with_mode(&mut writer, compress)?;
```
sydhds: Would use constant here with good naming

rln/src/circuit/mod.rs
```diff
@@ -121,22 +121,34 @@ pub struct ProofWithFlags(ArkProof<Curve>);
 impl CanonicalSerializeWithFlags for ProofWithFlags {
     fn serialize_with_flags<W: Write, F: Flags>(
         &self,
-        _writer: W,
-        _flags: F,
+        mut writer: W,
+        flags: F,
     ) -> Result<(), SerializationError> {
-        todo!()
+        self.serialize_compressed(&mut writer)?;
```
sydhds: After a quick verification, BIT_SIZE should not exceed 8 bits in arkwork repo so maybe for a first version set the same limit here? (see https://github.com/arkworks-rs/algebra/blob/4cec9f0e7a9a4f0d26433917980b1d4d5d8e521c/ff/src/fields/models/small_fp/serialize.rs#L16)

rln/src/circuit/mod.rs
```diff
@@ -121,22 +121,34 @@ pub struct ProofWithFlags(ArkProof<Curve>);
...
+        if F::BIT_SIZE > 0 {
+            writer.write_all(&[flags.u8_bitmask()])?;
```
sydhds: The flag is a bitmask and should set set with | (see https://github.com/arkworks-rs/algebra/blob/4cec9f0e7a9a4f0d26433917980b1d4d5d8e521c/ff/src/fields/models/small_fp/serialize.rs impl)

sydhds (changes requested):

rln/src/circuit/mod.rs
vinhtc27: Default::default() doesn't work here

rln/src/circuit/mod.rs
vinhtc27: I don't think we can use the |= here, SmallFp has spare bits because the prime is small relative to its storage (e.g. 31-bit in u128).

https://github.com/arkworks-rs/algebra/blob/4cec9f0e7a9a4f0d26433917980b1d4d5d8e521c/ff/src/fields/models/small_fp/small_fp_backend.rs#L20

A compressed Proof<Bn254> is (G1, G2, G1) every bit is occupied, OR-ing anything into the last byte corrupts the field value.

rln/src/circuit/mod.rs
```diff
@@ -434,4 +379,49 @@ mod test {
         .unwrap();
         assert!(matches!(err, GraphReadError::MaxOutMismatch { .. }));
     }
+
```
sydhds, line 382: Could these tests about serialization be in the serialize.rs file?

rln/src/protocol/mod.rs
```diff
@@ -8,6 +8,10 @@ mod slashing;
 mod witness;
 mod zk;
```
sydhds, line 10: Add comments for this constants. Could it be put in serialize.rs file?

sydhds (changes requested):

rln/src/circuit/mod.rs
vinhtc27: I'll move all serialization tests to the tests/serialize.rs in next PR (working on it rn so to avoid conflicting)

sydhds (approved):

PR #409, opened 2026-04-22
 refactor(PR 3): implement BE serialization for all new types

vinhtc27:


rln/src/protocol/proof.rs
```diff
@@ -212,9 +212,7 @@ pub fn rln_proof_values_to_bytes_le(rln_proof_values: &RLNProofValues) -> Vec<u8
             selector_used,
         } => {
             VERSION_BYTE_SIZE
```
sydhds, line 208: This is not "version" byte but rather "tag" or "variant" byte

rln/src/utils.rs
```diff
@@ -12,18 +12,31 @@ use rand::Rng;
...
+    // Get the modulus bit size of the base field
+    let modulus_bits: u32 = Fq::MODULUS_BIT_SIZE;
+    // Alignment boundary in bits for field element serialization
+    let alignment_bits: u32 = 64;
+    // Align to the next multiple of alignment_bits and convert to bytes
+    ((modulus_bits + alignment_bits - (modulus_bits % alignment_bits)) / 8) as usize
+};
+
 /// Byte size of the length prefix used when serializing variable-length vectors.
 pub const VEC_LEN_BYTE_SIZE: usize = 8;
```
sydhds: u64 size of?

rln/src/circuit/mod.rs
```diff
@@ -112,20 +118,67 @@ pub type Proof = ArkProof<Curve>;
 impl CanonicalSerializeBE for Proof {
     type Error = ProtocolError;
 
-    fn serialize<W: Write>(&self, _writer: W) -> Result<(), Self::Error> {
-        todo!()
+    fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
```
sydhds: Any reason to support Proof BE serialization or deserialization here? I thought we decided to only support LE for it?

rln/src/protocol/proof.rs
```diff
@@ -1095,84 +1138,54 @@ impl RecoverSecret<RLNProofValuesSingle> for RLNProofValuesMulti {
...
+        writer.write_all(&fr_to_bytes_be(&self.external_nullifier))?;
+        writer.write_all(&vec_fr_to_bytes_be(&self.ys))?;
+        writer.write_all(&vec_fr_to_bytes_be(&self.nullifiers))?;
+        writer.write_all(&vec_bool_to_bytes_be(&self.selector_used))?;
+        Ok(())
     }
 
     fn serialized_size(&self) -> usize {
-        todo!()
+        FR_BYTE_SIZE * (3 + self.ys.len() + self.nullifiers.len())
```
sydhds: Maybe split the computation per field (with comment) so it's clearer here?

rln/src/protocol/proof.rs
```diff
@@ -1095,84 +1138,54 @@ impl RecoverSecret<RLNProofValuesSingle> for RLNProofValuesMulti {
...
-    fn test_proof_values_v3_le_uncompressed_roundtrip() {
-        for pv in [make_proof_values_single(), make_proof_values_multi()] {
-            let mut buf = Vec::new();
-            pv.serialize_uncompressed(&mut buf).unwrap();
-            let deser = RLNProofValuesV3::deserialize_uncompressed(buf.as_slice()).unwrap();
-            assert_eq!(pv, deser);
-            assert_eq!(pv.uncompressed_size(), buf.len());
+    fn deserialize<R: Read>(mut reader: R) -> Result<Self, Self::Error> {
+        let mut bytes = Vec::new();
+        reader.read_to_end(&mut bytes)?;
```
sydhds: Should not use read_to_end here as you don't know if there is something in the buffer after the ProofValues no?

rln/src/protocol/proof.rs
```diff
@@ -1095,84 +1138,54 @@ impl RecoverSecret<RLNProofValuesSingle> for RLNProofValuesMulti {
...
+        let mut read = 0;
```
sydhds: Maybe we can consider using something like nom. it would be clearer to read 8 bytes, convert to Fr and avoid the `read += ...`.  Wdyt?

sydhds (changes requested):

rln/src/protocol/proof.rs
vinhtc27: This is just old code's constant, i'v added new `ENUM_TAG_SIZE` constant for the new refactor

rln/src/utils.rs
vinhtc27: What do u mean here?

rln/src/circuit/mod.rs
vinhtc27: I removed it an use the normal LE compressed format

rln/src/circuit/mod.rs
```diff
@@ -112,20 +112,22 @@ pub type Proof = ArkProof<Curve>;
 impl CanonicalSerializeBE for Proof {
```
sydhds: CanonicalSerializeBE for Proof is not needed here and should be removed completely :)

rln/src/circuit/mod.rs
```diff
@@ -112,20 +112,22 @@ pub type Proof = ArkProof<Curve>;
...
+        Ok(())
     }
 
     fn serialized_size(&self) -> usize {
-        todo!()
+        CanonicalSerialize::compressed_size(self)
     }
 }
 
 impl CanonicalDeserializeBE for Proof {
```
sydhds: same as previous

rln/src/utils.rs
sydhds: This mean that the vec len is using a u64 (serialized) for the size of a Vec no?

rln/src/utils.rs
```diff
@@ -494,9 +497,28 @@ impl IdSecret {
...
+        for (i, &limb) in bigint.0.iter().rev().enumerate() {
+            buf[i * 8..(i + 1) * 8].copy_from_slice(&limb.to_be_bytes());
+        }
+        let result = writer.write_all(buf.as_ref());
+        bigint.0.zeroize();
+        result
+    }
+
+    pub fn read_be<R: Read>(reader: &mut R) -> Result<Self, UtilsError> {
+        let mut buf = Zeroizing::new([0u8; FR_BYTE_SIZE]);
```
sydhds: Any reason to use Zeroizing here?

rln/src/utils.rs
```diff
@@ -494,9 +497,28 @@ impl IdSecret {
         Zeroizing::new(res)
     }
 
+    pub fn write_be<W: Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        let mut bigint = self.0.into_bigint();
+        let mut buf = Zeroizing::new([0u8; FR_BYTE_SIZE]);
+        for (i, &limb) in bigint.0.iter().rev().enumerate() {
```
sydhds: What is this for?

rln/src/utils.rs
```diff
@@ -556,3 +578,77 @@ impl From<IdSecret> for FrOrSecret {
         FrOrSecret::IdSecret(value)
     }
 }
+
+#[inline(always)]
```
sydhds: Maybe we should remove those inline. Not sure if this makes a real diff here no?

sydhds (changes requested):

rln/src/utils.rs
vinhtc27: It's using serialized `usize` for size of a Vec, on 32-bit systems, i'll be zero-padded to 8 bytes and directly represents the `usize` value on 64-bit systems.
This is the same way the old code handle Vec<T> serialization.

rln/src/utils.rs
vinhtc27: Previous impl `from_bytes_be` just borrows `&[u8]` so we doesn't have responsibility to zeroize.
And now new impl `read_be` owns the buffer so it must use Zeroizing

rln/src/utils.rs
vinhtc27: It's common pattern that these short utils function that use many time should inline, which all other old utils functions does too.

rln/src/utils.rs
vinhtc27: I updated these utils method for Fr with writer/reader pattern to avoid heap allocation, and use stack instead. And it fit better to those new `fn serialize<W: Write>(&self, mut writer: W)` and `fn deserialize<R: Read>(mut reader: R)` from new serialize traits. Some micro optimization but cleaner code too.

sydhds: Okay so now I understand and I think you should impl CanonicalSerialize (LE/BE) and CanonicalDeserialize trait instead of having this custom methods (read_be / write_be) wdyt?

rln/src/protocol/proof.rs
```diff
@@ -1041,20 +1062,42 @@ impl RecoverSecret<RLNProofValuesMulti> for RLNProofValuesSingle {
...
     }
 }
 
 impl CanonicalDeserializeBE for RLNProofValuesSingle {
     type Error = ProtocolError;
 
-    fn deserialize<R: Read>(_reader: R) -> Result<Self, Self::Error> {
-        todo!()
+    fn deserialize<R: Read>(mut reader: R) -> Result<Self, Self::Error> {
+        let mut buf = [0u8; FR_BYTE_SIZE];
```
sydhds: Cannot we read here in one call and fail early if it cannot be done?

sydhds (changes requested):

rln/src/utils.rs
vinhtc27: Great idea, it’ll be more consistent.

rln/src/protocol/proof.rs
vinhtc27: Yes, I already updated to reader pattern, deserialize each fields and fail early

rln/src/utils.rs
vinhtc27: I just tried to do so, but the code looks kind of ugly that way. It always has to call `CanonicalSerializeBE/CanonicalDeserializeBE` with the serialize/deserialize method for each field, and for those `Vec<T>` types, it looks even worse.

Let’s keep it clean like it is right now, I kind of prefer this way since it’s easier to read.

```rust
fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
    self.identity_secret.write_be(&mut writer)?;
    write_fr_be(&mut writer, &self.user_message_limit)?;
    write_vec_fr_be(&mut writer, &self.path_elements)?;
    write_vec_u8_be(&mut writer, &self.identity_path_index)?;
    write_fr_be(&mut writer, &self.x)?;
    write_fr_be(&mut writer, &self.external_nullifier)?;
    write_fr_be(&mut writer, &self.message_id)?;
    Ok(())
}    
```

sydhds (approved):

PR #410, opened 2026-04-27
refactor(PR 4): implement all stateless RLN features

vinhtc27:


rln/src/protocol/proof.rs
```diff
@@ -903,16 +906,89 @@ pub enum RLNProofValuesV3 {
 impl TryFrom<RLNWitnessInputV3> for RLNProofValuesV3 {
     type Error = ProtocolError;
 
-    fn try_from(_witness: RLNWitnessInputV3) -> Result<Self, Self::Error> {
-        todo!()
+    fn try_from(witness: RLNWitnessInputV3) -> Result<Self, Self::Error> {
```
sydhds: Each variant should have its own implementation (TryFrom<RLNWitnessInputSingle> for RLNProofValuesSingle for instance) and here we should just have a static dispatch

rln/src/protocol/proof.rs
```diff
@@ -925,27 +1001,33 @@ pub struct RLNProofValuesSingle {
...
-        todo!()
-    }
-}
-
 impl RecoverSecret for RLNProofValuesSingle {
     type Error = ProtocolError;
 
-    fn recover_secret(&self, _other: &Self) -> Result<Fr, Self::Error> {
-        todo!()
+    fn recover_secret(&self, other: &Self) -> Result<Fr, Self::Error> {
```
sydhds: recover_secret trait should return an IdSecret type here no?

rln/src/protocol/proof.rs
```diff
@@ -903,16 +906,89 @@ pub enum RLNProofValuesV3 {
...
 impl RecoverSecret for RLNProofValuesV3 {
     type Error = ProtocolError;
 
-    fn recover_secret(&self, _other: &Self) -> Result<Fr, Self::Error> {
-        todo!()
+    fn recover_secret(&self, other: &Self) -> Result<Fr, Self::Error> {
+        match (self, other) {
+            (RLNProofValuesV3::Single(s), RLNProofValuesV3::Single(o)) => s.recover_secret(o),
+            (RLNProofValuesV3::Multi(s), RLNProofValuesV3::Multi(o)) => s.recover_secret(o),
+            (RLNProofValuesV3::Single(s), RLNProofValuesV3::Multi(o)) => s.recover_secret(o),
```
sydhds: Should use | here to be explicit that order does not matter here no?

rln/src/protocol/witness.rs
```diff
@@ -939,36 +938,147 @@ pub enum RLNWitnessInputV3 {
...
+                    w.selector_used.len(),
+                    "max_out",
+                    graph.max_out,
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    pub(super) fn to_circuit_inputs(&self) -> Vec<(&'static str, Vec<FrOrSecret>)> {
```
sydhds, line 986: We should mark (add a comment) for  this to be reworked. I think we could do way better than creating this weird str -> Fr vec

rln/src/protocol/witness.rs
```diff
@@ -1043,13 +1265,23 @@ impl From<&RLNWitnessInputV3> for RLNPartialWitnessInputV3 {
 }
 
 impl From<&RLNWitnessInputSingle> for RLNPartialWitnessInputV3 {
```
sydhds, line 1289: Here we are forcing users to use From with a reference and paying the cost of a clone. Maybe we should provide both From impl (with ref & without ref) no?

rln/src/public.rs
```diff
@@ -815,36 +819,104 @@ impl<T: ZerokitMerkleTree, ZkProof> RLNV3<Stateful<T>, ZkProof> {
...
+        partial_proof: ZkProof::PartialProof,
+        witness: ZkProof::Witness,
+    ) -> Result<ZkProof::Proof, RLNError> {
+        Ok(self.zkp.finish_proof(partial_proof, witness)?)
+    }
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl RLNV3<Stateless, ArkGroth16Backend> {
+    /// Creates a stateless instance using the embedded Single circuit (tree_depth=20, max_out=1).
```
sydhds, line 865: Providing all of these "constructors" feels a bit weird here. Could we use a builder pattern here?

rln/src/public.rs
```diff
@@ -815,36 +819,104 @@ impl<T: ZerokitMerkleTree, ZkProof> RLNV3<Stateful<T>, ZkProof> {
...
+        proof: &Proof,
+        values: &RLNProofValuesV3,
+        x: &Fr,
+        roots: &[Fr],
+    ) -> Result<bool, RLNError> {
+        let verified = self.zkp.verify(proof, values)?;
+        if !verified {
+            return Err(VerifyError::InvalidProof.into());
+        }
+        if !roots.is_empty() && !roots.contains(&values.root()) {
```
sydhds, line 911: It might be more efficient to check for is_empty & cmp x values first? As it would early reject the verification no?

rln/src/bin/test_refactors.rs
```diff
@@ -1,24 +1,146 @@
...
-    let witness2 = RLNWitnessInputV3::Multi(RLNWitnessInputMulti::new()?);
-    let (proof2, proof_values2) = rln_stateful.generate_proof(witness2)?;
-    let ok = rln_stateful.verify_proof(&proof2, &proof_values2)?;
-    assert!(ok);
+    // Stateless Single
+    let backend = ArkGroth16Backend::new(zkey_single_v1().to_owned(), graph_single_v1().to_owned());
+    let rln = RLNV3::<Stateless, ArkGroth16Backend>::new(backend);
+
+    let (identity_secret, _) = keygen();
+    let witness_single1 = RLNWitnessInputV3::Single(RLNWitnessInputSingle::new(
```
sydhds: Here we must have From impl so we could simplify the code? Wdyt?

sydhds (changes requested):

rln/src/public.rs
vinhtc27: I'll add the builder pattern in PR8

rln/src/protocol/proof.rs
sydhds: Seems like this one has not been resolved?

vinhtc27: Right, i missed this one

sydhds (changes requested):

sydhds (approved):

PR #411, opened 2026-05-12
refactor(PR 5): implement missing stateless features and rln-wasm module

vinhtc27:
## Changes:
- Enhanced ArkGroth16Backend and RLNWitnessInputV3 with WASM support and new methods
- Added RLNProofV3 type, Result-returning V3 accessors, new error enum, WASM constructor, and generate_proof_from_calculated_witness method on RLNZkProof trait.
- Enhanced RLNProofV3 with serialization traits and updated related tests.
- Migrated WASM bindings to V3 types, and added truncate/extra byte tests for witness and proof values

### DIFF
--- a/rln/src/circuit/mod.rs
+++ b/rln/src/circuit/mod.rs
@@ -18,14 +18,11 @@ use ark_groth16::{
 use ark_relations::r1cs::ConstraintMatrices;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 
-#[cfg(not(target_arch = "wasm32"))]
-use self::error::GraphReadError;
-use self::error::ZKeyReadError;
-#[cfg(not(target_arch = "wasm32"))]
-use crate::circuit::iden3calc::{
-    graph::Node, storage::deserialize_witnesscalc_graph, InputSignalsInfo,
+use self::error::{GraphReadError, ZKeyReadError};
+use crate::{
+    circuit::iden3calc::{graph::Node, storage::deserialize_witnesscalc_graph, InputSignalsInfo},
+    partial_proof::PartialProof as ArkPartialProof,
 };
-use crate::partial_proof::PartialProof as ArkPartialProof;
 
 #[cfg(not(target_arch = "wasm32"))]
 const GRAPH_BYTES_SINGLE_V1: &[u8] = include_bytes!("../../resources/tree_depth_20/graph.bin");
@@ -120,7 +117,6 @@ pub type VerifyingKey = ArkVerifyingKey<Curve>;
 ///
 /// Contains the deserialized computation graph used for witness calculation.
 /// Parsing this once and reusing it avoids repeated deserialization overhead.
-#[cfg(not(target_arch = "wasm32"))]
 #[derive(Clone)]
 pub struct Graph {
     pub(crate) nodes: Vec<Node>,
@@ -142,7 +138,6 @@ pub fn zkey_from_raw(zkey_data: &[u8]) -> Result<Zkey, ZKeyReadError> {
 }
 
 /// Parses the witness calculator graph from raw bytes
-#[cfg(not(target_arch = "wasm32"))]
 pub fn graph_from_raw(
     graph_data: &[u8],
     expected_tree_depth: Option<usize>,
@@ -277,21 +272,23 @@ fn read_arkzkey_from_bytes_uncompressed(arkzkey_data: &[u8]) -> Result<Zkey, ZKe
 }
 
 #[derive(Clone)]
-pub struct ArkGroth16Backend {
+pub struct ArkGroth16BackendWithGraph {
     pub(crate) zkey: Zkey,
-    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) graph: Graph,
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-impl ArkGroth16Backend {
+impl ArkGroth16BackendWithGraph {
     pub fn new(zkey: Zkey, graph: Graph) -> Self {
         Self { zkey, graph }
     }
 }
 
-#[cfg(target_arch = "wasm32")]
-impl ArkGroth16Backend {
+#[derive(Clone)]
+pub struct ArkGroth16BackendWithoutGraph {
+    pub(crate) zkey: Zkey,
+}
+
+impl ArkGroth16BackendWithoutGraph {
     pub fn new(zkey: Zkey) -> Self {
         Self { zkey }
     }
--- a/rln/src/circuit/mod.rs
+++ b/rln/src/circuit/mod.rs
@@ -276,10 +276,10 @@ fn read_arkzkey_from_bytes_uncompressed(arkzkey_data: &[u8]) -> Result<Zkey, ZKe
     Ok(zkey)
 }
 
-#[cfg(not(target_arch = "wasm32"))]
 #[derive(Clone)]
 pub struct ArkGroth16Backend {
     pub(crate) zkey: Zkey,
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) graph: Graph,
 }
 
@@ -287,6 +287,13 @@ pub struct ArkGroth16Backend {
 impl ArkGroth16Backend {
     pub fn new(zkey: Zkey, graph: Graph) -> Self {
         Self { zkey, graph }
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+impl ArkGroth16Backend {
+    pub fn new(zkey: Zkey) -> Self {
+        Self { zkey }
     }
 }
 
--- a/rln/src/circuit/mod.rs
+++ b/rln/src/circuit/mod.rs
@@ -117,7 +117,7 @@ pub type VerifyingKey = ArkVerifyingKey<Curve>;
 ///
 /// Contains the deserialized computation graph used for witness calculation.
 /// Parsing this once and reusing it avoids repeated deserialization overhead.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Graph {
     pub(crate) nodes: Vec<Node>,
     pub(crate) signals: Vec<usize>,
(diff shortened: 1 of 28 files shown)


rln/src/protocol/proof.rs
```diff
@@ -142,52 +142,47 @@ impl RLNProofValues {
         &self.external_nullifier
     }
 
-    /// Returns the `y` value (only valid for SingleV1).
     pub fn y(&self) -> &Fr {
         match &self.outputs {
             RLNOutputs::SingleV1 { y, .. } => y,
             RLNOutputs::MultiV1 { .. } => {
-                panic!("y() is not available for MultiV1 proof values; use ys()")
+                todo!("y() is not available for MultiV1 proof values; use ys()")
```
sydhds, line 149: Why replacing panic!() with todo!() here?

rln/src/protocol/serialize.rs
```diff
@@ -477,6 +477,34 @@ impl CanonicalDeserializeBE for RLNPartialWitnessInputV3 {
     }
 }
 
+impl CanonicalSerializeBE for RLNProofV3 {
```
sydhds: This is misleading as here you are serializing Proof in LE format and not in BE. I think we already discuss many times that we do not want to provide SerializeBE for RLN proof no?

rln/src/protocol/witness.rs
```diff
@@ -938,6 +935,130 @@ pub enum RLNWitnessInputV3 {
...
+        match self {
+            Self::Multi(w) => Ok(&w.selector_used),
+            Self::Single(_) => Err(ProtocolError::FieldNotInVariant {
+                field: "selector_used",
+                variant: "Single",
+            }),
+        }
+    }
+
+    pub fn to_bigint_json(&self) -> Result<serde_json::Value, ProtocolError> {
```
sydhds, line 1013: What the idea behind this function? Where will it be used?

rln/src/protocol/zk.rs
```diff
@@ -78,17 +87,54 @@ impl RLNZkProof for ArkGroth16Backend {
...
+            let values = RLNProofValuesV3::try_from(witness)?;
+            Ok((proof, values))
+        }
+        #[cfg(target_arch = "wasm32")]
+        unreachable!(
+            "generate_proof requires a circuit graph; use generate_proof_with_witness on WASM instead"
+        )
+    }
+
+    fn generate_proof_with_witness(
```
sydhds: Not sure to understand the need for generate_proof_with_witness here (compared to generate_proof)?

sydhds (changes requested):

rln/src/protocol/serialize.rs
vinhtc27: Yes, it's still using serialize_compressed from CanonicalSerialize as LE internally, and only RLNProofValuesV3 is using CanonicalSerializeBE's method

rln/src/protocol/proof.rs
vinhtc27: I just searched the entire codebase to replace any panic!, unreachable!, or unwrap! usages in the main code.

The idea is to only allow expect in the core lib. Other panic-related macros should only be used in tests or for todo! / unimplemented! cases.

You can ignore this for now, I’ll clean up these old type impls in a future PR anyway.

rln/src/protocol/witness.rs
vinhtc27: It's already using in rln-wasm module, in wasm_rln.rs file
```rust
    #[wasm_bindgen(js_name = toBigIntJson)]
    pub fn to_bigint_json(&self) -> Result<Object, String> {
        let bigint_json = self.0.to_bigint_json().map_err(|err| err.to_string())?;

        let serializer = serde_wasm_bindgen::Serializer::json_compatible();
        let js_value = bigint_json
            .serialize(&serializer)
            .map_err(|err| err.to_string())?;

        js_value
            .dyn_into::<Object>()
            .map_err(|err| format!("{:#?}", err))
    }
```    
This function is input for the witness calculation step running separately in JS (you can find it in resources/witness_calculator.js file in the rln-wasm module).

rln/src/protocol/zk.rs
vinhtc27: This function takes the calculated_witness from the witness_calculator.js I mentioned above, instead of doing the witness calculation internally.

Normally, generate_proof also takes the witness and performs the witness calculation using the graph.bin file internally. This API exists for clients that need to perform the witness calculation separately, as mentioned above.

This is especially useful for client-side flows, where witness calculation can be done locally so private inputs do not need to be sent to the central prover. It's the same api we released in v1.0.0 and v2.0.0/

rln/src/protocol/serialize.rs
vinhtc27: WDYT? Should I rename it or just remove it? It’s still a nice wrapper for the FFI or WASM use case, since other languages can’t return tuples from functions like Rust can (e.g. -> (Proof, ProofValues)). So having a wrapper type like RLNProofV3 lets external users ignore those implementation details, which I think is okay.

sydhds: Implementing CanonicalSerializeBE is misleading here so I think we should remove it. We already have CanonicalSerialize avail? Do you think that someone will one a mixed serialization format (LE for proof, BE for proof values) ?

vinhtc27: I removed the wrapper type.

rln/src/protocol/zk.rs
```diff
@@ -1,56 +1,57 @@
...
-        + TryFrom<Self::Witness>
         + CanonicalSerialize
         + CanonicalDeserialize
         + CanonicalSerializeBE
         + CanonicalDeserializeBE;
     type Proof: CanonicalSerialize + CanonicalDeserialize;
     type Error;
 
-    fn generate_proof(
+    fn generate_proof_from_calculated_witness(
```
sydhds: Please use generate_proof here (trait RLNZkProof) and generate_proof_from_witness for trait RLNZkProofWithGraph)

rln/src/protocol/zk.rs
```diff
@@ -69,71 +70,64 @@ pub trait RLNPartialZkProof: RLNZkProof {
...
     type Values = RLNProofValuesV3;
     type Proof = Proof;
     type Error = RLNError;
 
-    fn generate_proof(
+    fn generate_proof_from_calculated_witness(
         &self,
-        witness: Self::Witness,
-    ) -> Result<(Self::Proof, Self::Values), Self::Error> {
+        calculated_witness: &[Fr],
```
sydhds: Why is the "calculated witness" type hardcoded here?

rln/src/public.rs
```diff
@@ -860,47 +874,25 @@ where
...
-impl<S> RLNV3<S, ArkGroth16Backend> {
-    /// Returns the message mode inferred from the loaded circuit.
-    pub fn message_mode(&self) -> MessageMode {
-        MessageMode::from(&self.zkp.graph)
-    }
-
-    /// Returns the maximum number of message-id slots for the loaded circuit.
-    pub fn max_out(&self) -> usize {
-        self.zkp.graph.max_out
+impl RLNV3<Stateless, ArkGroth16BackendWithoutGraph> {
```
sydhds: I would recommend to get one convention and stick to it: ArkGroth16Backend (without graph) ArkGroth16BackendWithGraph.

sydhds (changes requested):

rln/src/protocol/zk.rs
vinhtc27: I think it would be much clearer with this naming convention, as it distinguishes it from generate_proof_from_witness, which already a method of trait RLNZkProofWithGraph.

rln/src/protocol/zk.rs
```diff
@@ -1,56 +1,58 @@
...
+    protocol::{proof::RLNProofValuesV3, witness::RLNWitnessInputV3},
     utils::IdSecret,
 };
 
 pub trait RLNZkProof {
-    type Witness: CanonicalSerialize
-        + CanonicalDeserialize
-        + CanonicalSerializeBE
-        + CanonicalDeserializeBE;
+    type CalculatedWitness: AsRef<[Fr]>;
```
sydhds: I don't think we should have the AsRef<[Fr> trait bounds here and we should let the dev impl any type he wants / requires. 

Also I'm wondering if this should implement all the CanonicalSerialize traits here. Wdyt?

rln/src/protocol/zk.rs
sydhds: I think this really make sense as generate_proof is the baseline function (to generate a proof) and is used by generate_proof_from_witness and it would simplify the naming here. Wdyt?

rln/src/circuit/mod.rs
```diff
@@ -276,20 +271,29 @@ fn read_arkzkey_from_bytes_uncompressed(arkzkey_data: &[u8]) -> Result<Zkey, ZKe
...
 
-#[cfg(not(target_arch = "wasm32"))]
-impl ArkGroth16Backend {
+impl ArkGroth16BackendWithGraph {
     pub fn new(zkey: Zkey, graph: Graph) -> Self {
         Self { zkey, graph }
     }
 }
 
+#[derive(Clone)]
```
sydhds: Maybe also Debug?

sydhds (changes requested):

rln/src/protocol/zk.rs
vinhtc27: Good idea, it will make CalculatedWitness consistent with other types in these traits

rln/src/protocol/zk.rs
```diff
@@ -183,8 +176,49 @@ impl RLNPartialZkProof for ArkGroth16Backend {
...
             self.zkey.1.num_instance_variables,
             self.zkey.1.num_constraints,
-            full_assignment.as_slice(),
+            &calculated_witness,
         )?;
         Ok(proof)
     }
 }
+
+fn prove(zkey: &Zkey, calculated_witness: &[Fr]) -> Result<Proof, RLNError> {
```
sydhds: naming is a bit weird here. We are using prove function in ArkGroth16Backend::generate_proof. Can we have a consistent naming here (like verify function used in ArkGroth16Backend::verify)?

sydhds (changes requested):

rln/src/protocol/zk.rs
vinhtc27: Updated

sydhds (approved):

PR #413, opened 2026-05-18
chore(release): v2.0.2

### LINKED ISSUE #412: v2.0.1 nix flake isn't usable by downstream
Hi, from the logos-delivery side (we pin zerokit's flake, on v2.0.1).

v2.0.1's flake isn't usable by a downstream building from scratch — the cargoHash in nix/default.nix doesn't match what the build actually produces. We've worked around it on our end for now, but it'd be great to get a corrected cargoHash into a v2.0.x patch or future release. 

If this is already fixed in a newer release, feel free to close — just flagging since we're pinned to v2.0.1. Thanks!

seemenkina:
## Description

- **`fix(nix)`** — refresh `nix/default.nix` cargoHash. The v2.0.1 tag shipped with a stale hash; cold nix builds failed because the cache-served FOD masked the mismatch. Verified hash with logos-delivery's downstream override.
- **`feat(rln)`** — add `ffi_rln_proof_new` and `ffi_rln_proof_values_new` constructors. `FFI_RLNProof` and `FFI_RLNProofValues` exposed `_get_*` accessors for every field but no `_new`, so downstream consumers
that decompose a proof into a struct (e.g. waku-rln-relay) had to hand-assemble the 290-byte wire layout and round-trip through `ffi_bytes_le_to_rln_proof`. Closes that asymmetry.
- **`fix(rln)`** — drop a redundant `.into_iter()` in `pm_tree_adapter::set_range` that newer clippy flags as `useless_conversion`. Was blocking CI on default and `multi-message-id` builds.

`FFI_RLNPartialProof` was deliberately not given a `_new` — it has no field accessors either, so there's no symmetry gap to close (the only construction path is `ffi_bytes_le_to_rln_partial_proof`).

## Test plan

- [ ] `cargo check` on default / `--features stateless --no-default-features` / `--features multi-message-id`
- [ ] `cargo clippy --all-targets --tests --release -- -D warnings` clean on all three feature combinations
- [ ] `cargo test --release -p rln --test ffi` — covers 3 new tests (`test_ffi_rln_proof_new_roundtrip`, `test_ffi_rln_proof_new_invalid_groth16_bytes`, `test_ffi_rln_proof_values_new_roundtrip`)
- [ ] Cold `nix build .#rln` succeeds without a hash-mismatch error
- [ ] `cargo run --features headers --bin generate-headers` produces an `rln.h` containing the two new functions

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -134,12 +134,12 @@ jobs:
       - name: Build rln-wasm package
         run: |
           if [[ ${{ matrix.feature }} == "parallel" ]]; then
-            env CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=--shared-memory -C link-arg=--max-memory=1073741824 -C link-arg=--import-memory -C link-arg=--export=__wasm_init_tls -C link-arg=--export=__tls_size -C link-arg=--export=__tls_align -C link-arg=--export=__tls_base" \
+            env CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=--shared-memory -C link-arg=--max-memory=1073741824 -C link-arg=--import-memory -C link-arg=--export=__wasm_init_tls -C link-arg=--export=__tls_size -C link-arg=--export=__tls_align -C link-arg=--export=__tls_base -C link-arg=--export=__heap_base" \
             rustup run nightly wasm-pack build --release --target web --scope waku \
             --features parallel -Z build-std=panic_abort,std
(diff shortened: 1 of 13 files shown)


vinhtc27 (approved):
LGTM.
I thought you checked out directly from the master branch, which contains commits from two merged PRs, but you already continued the changes from the last v2 release.

vinhtc27 (changes requested):
Only one change I forgot to mention: it’s still missing the nightly build change, the same as the change in rln-wasm’s Makefile.

Those "link-arg=--export=__heap_base" params are already in master [here](https://github.com/vacp2p/zerokit/blob/c2e6b77fbe2838d5d7a76fc7d3e9243199ae067d/.github/workflows/nightly-release.yml#L119), but just in case you release this branch separately as v2.0.2.

seemenkina:
> I thought you checked out directly from the master branch, which contains commits from two merged PRs, but you already continued the changes from the last v2 release.

yep, the idea is to create patch for previous release without touching new architectural changes, so I only update nix and also add missing methods for the FFI

vinhtc27 (approved):

PR #414, opened 2026-05-23
refactor(PR 6): implement stateful features, remove witness calc from JS, add new serialize trait and partial proof to rln-wasm

vinhtc27:
## Changes:
- ~Bump arkworks 0.5.0 → 0.6.0 and migrate GR1CS API across circuit, partial proof, and zk modules.~
&rarr; New ark-ec 0.6.0 parallel MSM creates local Rayon thread pool, panics in WASM (can't spawn OS threads). 
&rarr; But wasm-bindgen-rayon only patches global pool, not local ones ([Failed CI on wasm parallel](https://github.com/vacp2p/zerokit/actions/runs/26329096884/job/77511945431))
- Implement all RLNV3<Stateful<T>, ZKP> tree methods on public.rs: set_leaf, get_merkle_proof, delete_leaf, init_tree_with_leaves, and more.
- Validate stateful end-to-end in test_refactors.rs using FullMerkleTree, OptimalMerkleTree, and PmTree in the same binary without feature flag switching.
- Remove ArkGroth16BackendWithGraph and replace JS witness calculation with native Rust via graph.bin.
- Change the generate_proof API back to return both Proof and ProofValues.
- Add new CanonicalSer/DeserializeMixed trait for the RLNProofV3 type.
- Add partial proof to rln-wasm and update example and benchmarks.


rln/src/public.rs
```diff
@@ -805,17 +803,102 @@ impl<T, ZkProof> RLNV3<Stateful<T>, ZkProof> {
...
+}
+
+impl<T, ZkProof> RLNV3<Stateful<T>, ZkProof>
+where
+    T: ZerokitMerkleTree,
+    T::Hasher: Hasher<Fr = Fr>,
+    T::Proof: ZerokitMerkleProof<Index = u8>,
+    <T::Proof as ZerokitMerkleProof>::Hasher: Hasher<Fr = Fr>,
+{
+    pub fn get_merkle_proof(&self, index: usize) -> Result<(Vec<Fr>, Vec<u8>), RLNError> {
```
sydhds: Why not returning a T::Proof here?

sydhds (changes requested):

rln/src/protocol/proof.rs
```diff
@@ -1014,6 +1025,14 @@ impl TryFrom<RLNWitnessInputSingle> for RLNProofValuesSingle {
     type Error = ProtocolError;
 
     fn try_from(w: RLNWitnessInputSingle) -> Result<Self, Self::Error> {
+        RLNProofValuesSingle::try_from(&w)
```
sydhds: TryFrom impl using try_from with ref ??

rln/src/protocol/zk.rs
```diff
@@ -28,34 +28,17 @@ pub trait RLNZkProof {
     type Proof: CanonicalSerialize + CanonicalDeserialize;
     type Error;
```
sydhds: I think we should have a bound to std::error::Error here wdyt?

rln/src/protocol/zk.rs
```diff
@@ -74,66 +57,64 @@ pub trait RLNPartialZkProof: RLNZkProofWithGraph {
...
+        Groth16::<_, CircomReduction>::create_proof_with_reduction_and_matrices(
+            &self.zkey.0,
+            r,
+            s,
+            &self.zkey.1,
+            self.zkey.1.num_instance_variables,
+            self.zkey.1.num_constraints,
+            &calculated_witness,
+        )
+        .map_err(ProtocolError::from)
```
sydhds: The double map_err here is a bit weird and looks like AI code? Maybe only one map_err call is enough here ?

rln/src/protocol/zk.rs
```diff
@@ -74,66 +57,64 @@ pub trait RLNPartialZkProof: RLNZkProofWithGraph {
...
     }
 
     fn verify(&self, proof: &Self::Proof, values: &Self::Values) -> Result<bool, Self::Error> {
-        verify(&self.zkey, proof, values)
+        let public_inputs: Vec<Fr> = match values {
+            RLNProofValuesV3::Single(v) => {
+                vec![v.y, v.root, v.nullifier, v.x, v.external_nullifier]
+            }
+            RLNProofValuesV3::Multi(v) => {
+                let mut inputs = Vec::with_capacity(3 * v.ys.len() + 3);
```
sydhds: Maybe make it more explicit and use len of nullifiers && len of selector_used here?

rln/src/protocol/proof.rs
```diff
@@ -1079,6 +1098,14 @@ impl TryFrom<RLNWitnessInputMulti> for RLNProofValuesMulti {
     type Error = ProtocolError;
 
     fn try_from(w: RLNWitnessInputMulti) -> Result<Self, Self::Error> {
+        RLNProofValuesMulti::try_from(&w)
```
sydhds: Same here as previous (using ref)?

rln-wasm/src/wasm_rln.rs
```diff
@@ -1,8 +1,7 @@
 #![cfg(target_arch = "wasm32")]
```
sydhds, line 1: Wondering if  WasmRLNWitnessInput::to_bigint_json is still useful here?

rln/src/protocol/zk.rs
```diff
@@ -28,34 +28,17 @@ pub trait RLNZkProof {
     type Proof: CanonicalSerialize + CanonicalDeserialize;
     type Error;
 
-    fn generate_proof(
-        &self,
-        calculated_witness: &Self::CalculatedWitness,
-    ) -> Result<Self::Proof, Self::Error>;
+    fn generate_proof(&self, witness: &Self::Witness) -> Result<Self::Proof, Self::Error>;
```
sydhds: We discussed that we'd rather return Proof & Proof values here no?

sydhds (changes requested):

rln/src/protocol/zk.rs
vinhtc27: I think we shouldn't do it. String doesn't impl it so we'd block valid backends if they try primitive types.

rln/src/protocol/zk.rs
vinhtc27: I’m still on the fence about the public API here. ProofValues doesn’t come from generate_proof, it’s derived directly from the witness (TryFrom<&RLNWitnessInput>). Returning it alongside the proof makes it look like it’s produced by proof generation, which isn’t really the case.

Also, C and JS can’t return tuples anyway, so if we want to expose both values through WASM/FFI we’d still need a serializable wrapper like RLNProof. But that gets messy quickly, since Proof only supports LE while ProofValues supports both BE and LE (this is going to confuse users).

We discussed this in the last PR as well, which is why I’m still leaning toward keeping the current approach.

rln/src/protocol/zk.rs
sydhds: Is that really needed here? I mean beside experimenting with the API? It feels a bit weird (at least to me) to not have at least a trait bound with std::error::Error. And with a call to .into() this works too...

vinhtc27: Just added!

rln/src/protocol/zk.rs
seemenkina: The endian asymmetry isn't a wart we're trying to hide — it's correct: ZK Proof is fixed-LE because Groth16 compressed form requires it, and ProofValues supports both because users genuinely need both for blockchain integration. Users never deserialize the ZK Proof half themselves; it's opaque bytes round-tripped back into verify. So bundling them in RLNProof exposes zero of the asymmetry to the caller. Splitting them, on the other hand, makes every caller responsible for keeping track of "Proof is LE / Values is whatever I picked" — which is exactly the confusion we're trying to avoid.

So if we split up it it will require amount of decision from end user:

- serialize each one separately
- pick an endianness for `values`
- remember which they picked when reading them back
- not lose the pairing
- somehow document to their own downstream that "proof is always LE, values is whatever I chose"

vinhtc27: I just changed the public API to return both, and I also added a new CanonicalSer/DeserializeMixed trait for the RLNProofV3 type.

rln/src/protocol/serialize.rs
```diff
@@ -664,3 +658,40 @@ impl CanonicalDeserializeBE for RLNProofValuesMulti {
         })
     }
 }
+
+pub trait CanonicalSerializeMixed: CanonicalSerialize {
+    type Error;
```
sydhds, line 667: CanonicalSerialize always return SerializationError. So maybe we could do the same here? This could apply to CanonicalSerializeBE as well wdyt?

rln/src/protocol/serialize.rs
```diff
@@ -664,3 +658,40 @@ impl CanonicalDeserializeBE for RLNProofValuesMulti {
         })
     }
 }
+
+pub trait CanonicalSerializeMixed: CanonicalSerialize {
```
sydhds, line 666: We should add some doc indicating why we need this trait

rln-wasm/src/wasm_rln.rs
```diff
@@ -99,9 +99,28 @@ impl WasmRLNProof {
...
-        let proof =
-            Proof::deserialize_compressed(&bytes.to_vec()[..]).map_err(|err| err.to_string())?;
-        Ok(WasmRLNProof(proof))
+        let rln_proof = RLNProofV3::deserialize_compressed(&bytes.to_vec()[..])
+            .map_err(|err| err.to_string())?;
+        Ok(WasmRLNProof(rln_proof))
+    }
+
+    #[wasm_bindgen(js_name = toBytesBE)]
+    pub fn to_bytes_be(&self) -> Result<Uint8Array, String> {
```
sydhds: the name here is misleading as we are using mixed endian... (same for from_bytes_be)...

sydhds (changes requested):

rln/src/protocol/serialize.rs
vinhtc27: I think we should keep type Error on both. CanonicalSerializeBE already returns UtilsError for primitives like Fr, IdSecret, Vec<Fr>, not SerializationError. And since CanonicalSerializeMixed calls into CanonicalSerializeBE internally, hardcoding SerializationError on either would force a lossy conversion and drop error info. 

I'll have dedicated PR later to simplify all these RLN's errors. But i think we should keep it like this for now.

rln-wasm/src/wasm_rln.rs
```diff
@@ -99,9 +99,28 @@ impl WasmRLNProof {
...
     pub fn from_bytes_le(bytes: &Uint8Array) -> Result<WasmRLNProof, String> {
-        let proof =
-            Proof::deserialize_compressed(&bytes.to_vec()[..]).map_err(|err| err.to_string())?;
-        Ok(WasmRLNProof(proof))
+        let rln_proof = RLNProofV3::deserialize_compressed(&bytes.to_vec()[..])
+            .map_err(|err| err.to_string())?;
+        Ok(WasmRLNProof(rln_proof))
+    }
+
+    #[wasm_bindgen(js_name = toBytesMixed)]
```
sydhds: for mixed functions, we should add some documentation as the name is not that informative

rln/src/protocol/serialize.rs
sydhds: THen maybe add TODO to indicate that Error refacto will be done

sydhds (changes requested):

sydhds (approved):

PR #416, opened 2026-05-30
refactor(PR 7): cleanup old error type, redesign str-keyed map with CalcWitness/CalcWitnessPartial traits

vinhtc27:


rln/src/protocol/serialize.rs
sydhds, line 1: * impl CanonicalSerializeBE for Fr -> error is UtilsError but serialize can only IOError 
  * CanonicalDeserializeBE can only return IOError & NonCanonicalFieldElement 

-> I think we should get rid of UtilsError (same for Vec<Fr/boo/u8) and have SerializeErrorV3 (with added variants) for all of these types (so it is consistent). 

Another point/
* trait CanonicalDeserializeMixed: CanonicalDeserialize + Sized 
  * is the supertrait Size needed here? If not, remove it, if needed this should be added to the other ser/der traits no?

rln/src/error.rs
```diff
@@ -123,3 +123,71 @@ pub enum RLNError {
     #[error("Verification error: {0}")]
     Verify(#[from] VerifyError),
 }
+
+// TODO(PR10): delete ProtocolError and RLNError; strip V3 suffixes
+
+/// Errors that can occur during RLN protocol operations (witness creation, proof generation, etc.)
+#[derive(Debug, thiserror::Error)]
+pub enum ProtocolErrorV3 {
```
sydhds: While having a big enum for all errors is convenient. I'm thinking this is not convenient to use for developer using it. 
I think we should group them by kind of operation wdyt?

rln/src/error.rs
```diff
@@ -123,3 +123,71 @@ pub enum RLNError {
...
+
+/// Errors that can occur during serialization/deserialization of RLN types.
+#[derive(Debug, thiserror::Error)]
+pub enum SerializeErrorV3 {
+    #[error("IO error: {0}")]
+    Io(#[from] Error),
+    #[error("RLN utility error: {0}")]
+    Utils(#[from] UtilsError),
+    #[error("Arkworks serialization error: {0}")]
+    Arkworks(#[from] SerializationError),
```
sydhds: Variant name is not very informative no?

rln/src/error.rs
```diff
@@ -123,3 +123,71 @@ pub enum RLNError {
...
+}
+
+/// Top-level error type for all V3 RLN operations
+#[derive(Debug, thiserror::Error)]
+pub enum RLNErrorV3 {
```
sydhds: I might be wrong here but I feel that the following variants:
- Zkey / Graph / Config 

are only initialize variants? If yes, can we regroup them so a developer would easily understand where these error variants could come from?

rln/src/protocol/witness.rs
```diff
@@ -1122,6 +1067,127 @@ impl From<RLNWitnessInputMulti> for RLNWitnessInputV3 {
     }
 }
 
+impl CalcWitness for RLNWitnessInputV3 {
+    fn calc_witness(
+        &self,
+        graph: &Graph,
+    ) -> Result<Vec<Fr>, crate::circuit::error::WitnessCalcError> {
+        let inputs: Vec<(String, Vec<FrOrSecret>)> = match self {
```
sydhds, line 1059: so here, we just translate a Witness (struct of Fr) to a Vec<(String, Vec<FrOrSecret>)>. My idea was more to check if we can get rid of these translation and maybe use the Witness struct directly.

sydhds (changes requested):

rln/src/protocol/witness.rs
vinhtc27: I tried a few ways to get rid of it, but doing so still requires breaking the iden3calc functions into multiple smaller helper functions and moving them into the protocol module, which actually looks even worse.

I also tried creating functions that take Witness directly, but that would require importing a type from the outer module into the inner module, which isn't ideal.


Btw, I don't think we should change the iden3calc module either, because the code was copied from [circom-witnesscalc](https://github.com/iden3/circom-witnesscalc) repo.

rln/src/error.rs
```diff
@@ -123,3 +123,96 @@ pub enum RLNError {
...
+    },
+}
+
+/// Errors that can occur during serialization and deserialization of RLN types
+#[derive(Debug, thiserror::Error)]
+pub enum SerializationErrorV3 {
+    #[error("IO error: {0}")]
+    Io(#[from] Error),
+    #[error("Arkworks canonical serialization error: {0}")]
+    ArkworksSerializationError(#[from] SerializationError),
```
sydhds: Maybe the variant name could be just: "Serialize" and the error msg is already mentionning that it comes from Arkworks crate wdyt?

rln/src/protocol/serialize.rs
```diff
@@ -49,6 +49,17 @@ pub const FR_LIMB_BYTE_SIZE: usize = 8;
...
+fn serialize_usize_be(input: usize) -> [u8; VEC_LEN_BYTE_SIZE] {
+    let mut bytes = [0u8; VEC_LEN_BYTE_SIZE];
+    let input_bytes = input.to_be_bytes();
+    let offset = VEC_LEN_BYTE_SIZE - input_bytes.len();
+    bytes[offset..].copy_from_slice(&input_bytes);
+    bytes
+}
+
 pub trait CanonicalSerializeBE {
     type Error;
```
sydhds: Add std::error::Error bound (same for other ser/der traits)?

rln/src/error.rs
```diff
@@ -1,4 +1,4 @@
-use std::{array::TryFromSliceError, num::TryFromIntError};
+use std::{array::TryFromSliceError, io::Error, num::TryFromIntError};
```
sydhds: Minor things but to be consistent in the code:
* Error -> std::error::Error
* io::Error -> std::io::Error 

This will help when reading the code

rln/src/error.rs
sydhds, line 1: I'm still thinking the current error design is not helpful for developers and I would recommend this: 

* for RLNZkProof trait, error type is to generate_proof method
* add a new error type for verify method

* for trait RLNPartialZkProof, would implement the same things (another error for generate_partial & another error for finish_proof)

For witness:
* Add custom error per witness new (RLNWitnessInputSingle, ...) 
* Get rid of error for missing field, impl getter on each variant and return None if the field does not exist in enum 

* For RLNv3:
  * Return tree error for tree operation (not RLNErrorV3)
  * Return the error returned by generate_proof OR verify  

I think this would avoid this big enum errors and would be easier to deal with for Zerokit dev user. Open to discussion on this matter if needed

sydhds (changes requested):

rln/src/protocol/witness.rs
```diff
@@ -977,139 +984,64 @@ impl RLNWitnessInputV3 {
...
-            }),
+            Self::Multi(w) => Some(&w.selector_used),
+            Self::Single(_) => None,
         }
     }
 }
 
 impl RLNWitnessInputV3 {
-    pub(super) fn validate_against_graph(&self, graph: &Graph) -> Result<(), ProtocolError> {
+    pub(super) fn validate_against_graph(&self, graph: &Graph) -> Result<(), GenerateProofError> {
```
sydhds, line 1010: Maybe create a new error because here the name is a bit misleading no?

sydhds (changes requested):

rln/src/protocol/witness.rs
vinhtc27: This method is only used for proof generation, so I don't think it needs a separate error type. Also, if we introduce a new error, we'd either have to wrap it in GenerateProofError so it can be used by the RLNZkProof trait (which is what we're trying to avoid), or add another associated error type to RLNZkProof specifically for this case (it already has GenerateProofError and VerifyProofError). Both options seem unnecessarily complicated, so I decided to keep it this way.

sydhds (approved):

PR #417, opened 2026-06-05
refactor(PR 8): implement Builder pattern for RLNV3 type using external crate (bon) and update rln-cli module

vinhtc27:
Changes:
- Implemented Builder pattern for RLNV3 types using bon crate
- Updated rln-cli module to use RLNV3 types
- Update RFC links


rln/src/builder.rs
sydhds, line 1: I would have expected a more explicit builder with something like: 

```
let rln_stateless = RLNBuilder::build()
    .stateless()
    .graph(graph_data) # optional - use default if not provided
    .zkey(zkey_data); # optional - use default if not provided

let rln_stateless = RLNBuilder::build()
    .statefull()
    .tree_depth(depth)
   .tree_config(); # optional 
```

Wdyt?

sydhds (changes requested):

rln/src/builder.rs
vinhtc27: Hey, after implementing the FFI module, I think the builder pattern is kind of redundant (at least in its current form).

It doesn't really make things any easier for either native Rust or FFI. The FFI implementation can already be created without the builder (what I mean is that it makes no difference), and if I try to make the builder more explicit, as you suggested, it actually makes the FFI API harder to use. 

I also tried moving the builder into the FFI API, but that would require passing pointers in and out of multiple functions, which is very error-prone.

Maybe we shouldn't expose the builder through FFI at all (like not even use it in Rust's FFI files, just use normal init method) and just keep it as a native Rust API. That way, we can keep the FFI side simple and easy to integrate with from other languages, while still having a more idiomatic builder pattern for native Rust.

WDYT?

rln/src/pm_tree_adapter.rs
sydhds, line 1: I think we should try to use the bon crate for the PmtreeConfig. It will reduce the code that we have to maintain here

rln-wasm/src/wasm_rln.rs
```diff
@@ -19,11 +19,10 @@ impl WasmRLN {
...
     ) -> Result<WasmRLN, String> {
-        let rln = RLNV3::<Stateless, ArkGroth16Backend>::new_with_params(
-            zkey_data.to_vec(),
-            graph_data.to_vec(),
-        )
-        .map_err(|err| err.to_string())?;
+        let graph =
+            graph_from_raw(&graph_data.to_vec(), None, None).map_err(|err| err.to_string())?;
+        let zkey = zkey_from_raw(&zkey_data.to_vec()).map_err(|err| err.to_string())?;
+        let rln = RLNBuilder::stateless().graph(graph).zkey(zkey).build();
```
sydhds: I'm thinking if we could also provide something like: 

```
let rln = RLNBuilder::stateless().with_graph_file(graph_data).with_zkey_file(zkey_data).build();
```

And maybe (maybe this would be only for Rust) 

```
// Use default graph / zkey integrated in the lib
let rln = RLNBuilder::stateless().default(). build();
```

Note: the code `let rln = RLNBuilder::stateless().graph(graph).zkey(zkey).build();` should still be provided

rln/tests/serialize.rs
```diff
@@ -263,44 +263,40 @@ mod test {
...
-                Fr::from(10u64),
-                vec![Fr::from(1u64), Fr::from(2u64)],
-                vec![0u8, 1u8],
-                Fr::from(5u64),
-                Fr::from(7u64),
-                Fr::from(3u64),
-            )
-            .unwrap(),
-        )
+        RLNWitnessInputV3::new_single()
```
sydhds, line 266: Can we have something more explicit / shorter: 
* user_message_limit -> message_limit
* path_elements / identity_path_index: have an alternative that accept a T: MerkleTreeProof
  * Note that both way should exist (specifying path_elements & path_index OR merkle tree proof) 
  * At least a TODO for the function accepting a MerkleTreeProof
* x -> signal / message (and group with message_id)
* external_nullifier -> nullifier

Also a TODO for integrating the validate_against_graph in the builder?

The builder for the multi message id should have the same modifications as well...

rln-cli/src/examples/stateless.rs
```diff
@@ -155,37 +157,39 @@ impl RLNSystem {
...
-            .insert(*proof_values.nullifier(), proof_values);
-        println!("Message verified and accepted");
         Ok(())
     }
 
     fn handle_duplicate_nullifier(
         &mut self,
-        previous_proof_values: RLNProofValues,
-        current_proof_values: RLNProofValues,
+        previous_proof_values: RLNProofValuesV3,
```
sydhds: Maybe use ref here?

sydhds (changes requested):

rln-wasm/src/wasm_rln.rs
vinhtc27: The file handling should be in the NodeJS/browser side, but i'll add the default option for the native Rust

rln/src/pm_tree_adapter.rs
vinhtc27: Already use it

rln/tests/serialize.rs
vinhtc27: I added the TODO for PR 11, after the cleanup, and for the `user_message_limit` field, i think we should keep it the same as the RFC so we dont confuse user.

sydhds: So maybe we can have the same field name as the RFC and also shortcuts wdyt?

rln-cli/src/examples/multi_message_id.rs
```diff
@@ -60,39 +61,33 @@ impl Identity {
...
-        let tree_config = PmtreeConfigBuilder::new()
-            .path("./database")
-            .temporary(false)
-            .cache_capacity(1073741824)
-            .flush_every_ms(500)
-            .mode(Mode::HighThroughput)
-            .use_compression(false)
-            .build()?;
-        let rln = RLN::new_with_params(
+        let pm_tree_config: PmTreeConfig = r#"{
```
sydhds, line 71: Why not using the pmtree config builder anymore?

rln-cli/src/examples/relay.rs
```diff
@@ -139,63 +149,65 @@ impl RLNSystem {
...
         Ok(proof_values)
     }
 
-    fn check_nullifier(&mut self, proof_values: RLNProofValues) -> Result<()> {
-        if let Some(previous_proof_values) = self.used_nullifiers.get(proof_values.nullifier()) {
-            self.handle_duplicate_nullifier(previous_proof_values.clone(), proof_values)?;
-            return Ok(());
+    fn check_nullifier(&mut self, proof_values: RLNProofValuesV3) -> Result<()> {
+        if let Some(nullifier) = proof_values.nullifier() {
+            if let Some(previous_proof_values) = self.used_nullifiers.get(&nullifier).cloned() {
```
sydhds, line 187: This line could be simplified in one line no?

sydhds (changes requested):

rln-cli/src/examples/multi_message_id.rs
vinhtc27: I used difference tree, for each example, PmTreeConfig builder is in relay.rs example.
This example demonstrates the json config parse to PmTreeConfig.

rln-cli/src/examples/relay.rs
vinhtc27: Yes, something like this
```rust
   if let Some(nullifier) = proof_values.nullifier()
        && let Some(previous_proof_values) = self.used_nullifiers.get(&nullifier).cloned()
    {
```
but this only allowed in the Rust 2024 edition, but we're using the 2021 edition. I tried upgrading to the 2024 edition before, but some of the external crates we depend on produced errors due to the edition change.

rln/tests/serialize.rs
vinhtc27: I prefer explicit name for method and field name, and also i just tried to use rename macro from bon, it doesn't allow 2 name for the same method so we need another function for shorter name. 
Let's not duplicate the inner validation of RLNWitnessInput then, i think the code is pretty clean already.

sydhds (approved):

PR #418, opened 2026-06-10
refactor(PR 9): new V3 impl for FFI, migrate FFI/WASM utils to serialize trait and split and standardize C/Nim/JS examples

vinhtc27:


rln/ffi_c_examples/main.c
sydhds, line 1: This example is really huge and very difficult to reason about C API usage. Would it be the time to split this example in multiple ones?
It also seems to me that we should have unit tests here for the C ffi API & and these tests should be run in the CI wdyt?

sydhds (changes requested):

rln/ffi_c_examples/main.c
vinhtc27: I'll split the example into multiple ones. For CI, though, I don't think we need it because it adds too much complexity from other languages, and we already have plenty of FFI test cases.

rln-wasm/examples/0_common.js
```diff
@@ -0,0 +1,168 @@
...
+export const TREE_DEPTH = 20;
+export const MAX_OUT = 4;
+
+export function debugUint8Array(uint8Array) {
+  return Array.from(uint8Array, (byte) =>
+    byte.toString(16).padStart(2, "0"),
+  ).join(", ");
+}
+
+export async function initRLN(multiMessageId = false) {
```
sydhds: Maybe make the argument a bit more explicit like: 
* enableMultiMessageId 
* enableMultiMessageIdFeature
?

rln-wasm/examples/common.js
sydhds, line 1: Should rename 0_common.js -> common.js otherwise the user view the 0_common as a possible example (and it's not)

rln-wasm/examples/0_common.js
```diff
@@ -0,0 +1,168 @@
...
+  console.log("  - RLN instance created successfully");
+  console.log("  - circuit tree depth = " + TREE_DEPTH);
+  if (multiMessageId) {
+    console.log("  - circuit max out = " + MAX_OUT);
+  }
+
+  return { rlnWasm, rlnInstance, multiMessageId };
+}
+
+export function createMember(env) {
```
sydhds: env?

rln-wasm/examples/0_common.js
```diff
@@ -0,0 +1,168 @@
...
+      computedRoot,
+      defaultHashes[i - 1],
+    );
+  }
+  console.log("  - computed root = " + computedRoot.debug());
+
+  const roots = rlnWasm.VecWasmFr.new();
+  roots.push(computedRoot);
+
+  return { pathElements, identityPathIndex, computedRoot, roots };
```
sydhds: There is no MerkleProof type in Wasm? Also it's a bit unclear why we return computedRoot & roots here...

rln-wasm/examples/0_common.js
```diff
@@ -0,0 +1,168 @@
...
+  return {
+    identity,
+    identitySecret,
+    idCommitment,
+    userMessageLimit,
+    rateCommitment,
+  };
+}
+
+export function buildMerkleProof(env, rateCommitment) {
```
sydhds: buildMerkleProof -> computeMerkleProof?

rln-wasm/examples/2_serialization.js
sydhds, line 1: This example looks like more unit test than a real example. Do we really need to serialize everything (Proof, PartialProof, witness, partialWitness) here? I think one straight example with witness / proof is enough here no?

rln-wasm/examples/recover_secret.js
sydhds, line 1: This example should be renamed to "recover secret" no?

sydhds (changes requested):

rln-wasm/examples/0_common.js
vinhtc27: I dropped computedRoot, but roots is still needed: the proof has to be verified against the root of the tree that the merkle proof (path_elements, path_index) came from, so we pass it to verifyWithRoots to check against the correct root.

rln/ffi_nim_examples/README.md
sydhds, line 1: There is no way in Nim to provide struct (or class) with a drop implementation (that will automatically free the underlying object)? Tthat would give a better api no?

rln/ffi_c_examples/README.md
sydhds, line 1: Two things that we could improve here:
* I think most of the examples could leak memory if they stop early (as we free all the objects at the end of the program)
* Can we rename or add type alias. Some type names are really really long and looks scary wdyt?

sydhds (changes requested):

rln/ffi_nim_examples/README.md
vinhtc27: I think we should kept the examples using the raw C ABI on purpose, the explicit ffi_*_free calls make the ownership contract crystal clear.

rln/ffi_c_examples/README.md
vinhtc27: As for the memory leak concern, I checked all of the examples with valgrind and leaks tool to make sure they're safe. For mid crash case, I tried adding goto cleanup paths in the C examples, but I think it's better to keep them simple and raw.

The examples are meant to demonstrate the basic ownership model, and users should decide whatever RAII pattern fits their language and coding style.

And for the horible long type name, I use alias for all of it in both examples, its look much cleaner now.

sydhds: Ok for keeping the example as simple as possible but maybe add a small node about that?

rln/ffi_nim_examples/README.md
sydhds: But think as a Nim dev that would like to use the RLN api and see the "ugly" C API everywhere. If I were in this position, I would rather like to have (or read) idiomatic Nim code (I assume that Nim dev rarely takes care of memory freeing, same as Rust dev but I might be wrong here)

rln/ffi_c_examples/README.md
vinhtc27: Yes i added a note in both C and Nim examples README files: https://github.com/vacp2p/zerokit/blob/refactor/pr9-ffi-v3/rln/ffi_c_examples/README.md#memory-ownership

rln/ffi_nim_examples/README.md
vinhtc27: Hmm, this is kinda splitted examples from the original Nim file which Ekaterina wroted before.
I didn't really code Nim, like these are all my exposure to Nim so not sure the idiomatic way for doing this.
I will add her to this PR to ask for review of these Nim examples.

sydhds (approved):

PR #419, opened 2026-06-13
refactor(PR 10): migrate all modules and testcases to new v3 types and remove old code

vinhtc27:
# Changes

## Public API

- Strip `V3` suffix from all types
- Add `RLN::verify_with_signal(proof, values, x)`
- Replace old `RLN` struct with new type-state `RLN<State, ZKP>` + `RLNBuilder`
- API consistent across Rust / WASM / FFI: `verify` (pure ZK), `verify_with_signal`, `verify_with_roots`

## Tree feature flag cleanup

- Drop `fullmerkletree`, `optimalmerkletree`, `pmtree-ft` features
- Gate: `feature = "pmtree-ft"` → `not(target_arch = "wasm32")` for /utils
- Delete `poseidon_tree.rs` and feature-mutex `compile_error!` feature guards

## Delete utils.rs

- Delete entire `rln/src/utils.rs`
- Move `IdSecret` + `FrOrSecret` into `rln/src/circuit/mod.rs`
- Delete redundant util fns + old error types (`UtilsError`, `ProtocolError`, `RLNError`)

## Cleanup rln-utils module

- `sled` now optional: `pmtree-ft = ["pmtree", "sled"]`
- `Fr::zero()` → `Fr::ZERO` (drop `num-traits`)

## Bug fixes

- Function `hash_to_field_be` was identical to `_le` due to `hash.reverse()` before BE parse → switched to `Fr::from_be_bytes_mod_order`
- `hash_to_field_le` switched to `Fr::from_le_bytes_mod_order` (drops `BigUint` heap detour)
- `graph_from_raw` depth mismatch returns `GraphReadError::MissingSignal`

## Test migration

- Rewrote integration tests under `rln/tests/` with naming convention `test_<action>_<scope>_<variant>`:
- Deleted `tests/protocol.rs`, `tests/utils.rs`, `tests/poseidon_tree.rs`


.github/workflows/ci.yml
```diff
@@ -22,129 +22,95 @@ on:
...
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 60
-
-    name: Test - ${{ matrix.crate }} - ${{ matrix.platform }} - ${{ matrix.feature }}
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout sources
+        uses: actions/checkout@v5
       - name: Install nightly toolchain
         uses: dtolnay/rust-toolchain@nightly
```
sydhds, line 89: Do we need nightly rust for compiling Zerokit? Maybe we could add a rust-toolchain.md to fix a Rust version then?

utils/tests/poseidon_constants.rs
```diff
@@ -16,20 +16,17 @@ mod test {
...
-
-        // We remove any quote present and we trim
-        let single_quote: char = '\"';
-        let mut input_clean = input.replace(single_quote, "");
-        input_clean = input_clean.trim().to_string();
+    fn fr_from_hex(input: &str) -> Fr {
+        input
+            .trim_start_matches("0x")
+            .chars()
+            .fold(Fr::from(0), |acc, c| {
```
sydhds: Could we keep the BigUint::from_str_radix code here as it is more explicit and readable?

rln/Cargo.toml
sydhds, line 1: There are a few dependencies (rust crate) that we could update like bon, rand, zeroize (non exhaustive list here)...

rln/tests/keygen.rs
```diff
@@ -0,0 +1,108 @@
...
+                acc * Fr::from(16) + Fr::from(c.to_digit(16).unwrap())
+            })
+    }
+
+    #[test]
+    fn test_keygen_commitment_relation() {
+        let (identity_secret, id_commitment) = keygen();
+        let mut to_hash = [*identity_secret.clone()];
+        let expected_id_commitment = poseidon_hash(&to_hash);
+        to_hash[0].zeroize();
```
sydhds, line 13: Any reason to zeroize in a unit test?

rln/tests/keygen.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        // We check again if the identity pair generated with the same seed phrase corresponds to the previously generated one
+        let (identity_secret, id_commitment) = seeded_keygen(seed_phrase.as_bytes());
+
+        assert_eq!(identity_secret, expected_identity_secret_seed_phrase);
+        assert_eq!(id_commitment, expected_id_commitment_seed_phrase);
+    }
+
+    #[test]
+    fn test_extended_seeded_keygen_hex_fixture() {
```
sydhds: Can we regroup the extended_seed_keygen tests (like for the test_seed_keygen test)?

rln/src/error.rs
```diff
@@ -134,7 +35,7 @@ pub enum SerializationErrorV3 {
```
sydhds, line 35: Can we get rid of the "RLN" prefix here?

rln/src/protocol/zk.rs
```diff
@@ -67,18 +66,18 @@ pub trait RLNPartialZkProof: RLNZkProof {
 }
```
sydhds, line 67: Maybe we can get rid of "RLN" prefix here wdyt?

rln/tests/ffi_utils.rs
```diff
@@ -3,8 +3,15 @@ mod test {
     use rand::Rng;
     use rln::{ffi::ffi_utils::*, prelude::*};
 
+    fn fr_from_hex(hex: &str) -> Fr {
+        hex.trim_start_matches("0x")
+            .chars()
+            .fold(Fr::from(0), |acc, c| {
```
sydhds: Same as previous comment but please use BigUint function here instead of rolling your own code...

rln/tests/witness.rs
sydhds, line 1: Can we split this file between: single, multi, partial? Would be easier to read and review

rln/tests/public.rs
```diff
@@ -1,15 +1,26 @@
...
+
+    const NO_OF_LEAVES: usize = 256;
+
+    type StatefulRLN = RLN<Stateful<PmTree>, ArkGroth16Backend>;
 
     fn fq_from_str(s: &str) -> Fq {
         Fq::from_str(s).unwrap()
     }
 
+    fn fr_from_dec(s: &str) -> Fr {
```
sydhds: Can't we use the Fr::from_str function diirectly instead of having this func that have little interest?

sydhds (changes requested):
2 things: 
* I'm wondering if this PR should be merged in master. Until API 3.0.0 is ready, maybe we should merge in develop / zerokit_3_0 branch then merge at the end in master?
* Please add a TODO for verify_with_signal / verify_with_roots methods. Name is a bit confusing and also confusing in terms of dev UX (but we should fix this in another PR)

rln/Cargo.toml
vinhtc27: I'll do a final bump in PR12 before release

.github/workflows/ci.yml
vinhtc27: Yes we do need the nightly toolchain for the parallel feature on wasm.

rln/tests/witness.rs
vinhtc27: I mean this test file is pretty short, about only 350 lines tiny compare to other one.
I also arranged the tests to cover each valid and error case, following the same order as error, so reading this should be easy.

rln/tests/keygen.rs
vinhtc27: I just migrated it from the old test file. It’s still correct usage, right?

vinhtc27:
> * I'm wondering if this PR should be merged in master. Until API 3.0.0 is ready, maybe we should merge in develop / zerokit_3_0 branch then merge at the end in master?

I don’t think anyone is using the code directly from master anyway, only tagged versions. 
Also, the last 9 PRs have already been merged into master already.

rln/src/error.rs
vinhtc27: Nice catch!

rln/src/protocol/zk.rs
vinhtc27: I think we should keep the convention the same here, for the least rename when migration to V3.

.github/workflows/ci.yml
sydhds: So maybe add a rust-toolchain.md to specify a (working) nightly version ?

rln/tests/keygen.rs
sydhds: Right but this is a test so not needed? If you think this can be used as an example, maybe add a comment?

rln/src/protocol/zk.rs
sydhds: Not sure to understand here :-/

sydhds:
> > * I'm wondering if this PR should be merged in master. Until API 3.0.0 is ready, maybe we should merge in develop / zerokit_3_0 branch then merge at the end in master?
> 
> I don’t think anyone is using the code directly from master anyway, only tagged versions. Also, the last 9 PRs have already been merged into master already.

Previous PR did not remove the previous API (only added the V3) but here you remove the old APIs. Also we could not assume anyone use the zerokit from git so that's why I suggest a branch here.

sydhds (changes requested):

rln/src/protocol/zk.rs
vinhtc27: I mean it should be the same convention as other types defined in this protocol folder (like RLNWitnessInput, RLNProofValues,..)

.github/workflows/ci.yml
vinhtc27: I’d prefer to keep using the latest nightly version here. In the past, the nightly CI has caught several issues caused by WASM updates, whether it was browser-related problems, `wasm-pack` errors, or deprecated Node.js versions used for building WASM. 

Since the WASM thread pool implementation keeps changing, these kinds of issues can show up at any time, and latest nightly toolchain can catch it. 

We still have stable builds available through tagged [GitHub releases](https://github.com/vacp2p/zerokit/releases/tag/v2.0.2) and published [npm packages](https://www.npmjs.com/package/@waku/zerokit-rln-wasm-parallel), so users who want a stable version can continue using those.

vinhtc27:
@sydhds 

> Previous PR did not remove the previous API (only added the V3) but here you remove the old APIs. Also we could not assume anyone use the zerokit from git so that's why I suggest a branch here.

I just closed this PR. Let's continue the discussion in the https://github.com/vacp2p/zerokit/pull/420 PR, which merge to the [develop/v3.0.0](https://github.com/vacp2p/zerokit/tree/develop/v3.0.0) branch.

PR #420, opened 2026-06-16
refactor(PR 10): migrate all modules and testcases to new v3 types and remove old code

vinhtc27:
# Changes:

## Public API

- Strip `V3` suffix from all types
- Add `RLN::verify_with_signal(proof, values, x)`
- Replace old `RLN` struct with new type-state `RLN<State, ZKP>` + `RLNBuilder`
- API consistent across Rust / WASM / FFI: `verify` (pure ZK), `verify_with_signal`, `verify_with_roots`

## Tree feature flag cleanup

- Drop `fullmerkletree`, `optimalmerkletree`, `pmtree-ft` features
- Gate: `feature = "pmtree-ft"` → `not(target_arch = "wasm32")` for /utils
- Delete `poseidon_tree.rs` and feature-mutex `compile_error!` feature guards

## Delete utils.rs

- Delete entire `rln/src/utils.rs`
- Move `IdSecret` + `FrOrSecret` into `rln/src/circuit/mod.rs`
- Delete redundant util fns + old error types (`UtilsError`, `ProtocolError`, `RLNError`)

## Cleanup rln-utils module

- `sled` now optional: `pmtree-ft = ["pmtree", "sled"]`
- `Fr::zero()` → `Fr::ZERO` (drop `num-traits`)

## Bug fixes

- Function `hash_to_field_be` was identical to `_le` due to `hash.reverse()` before BE parse → switched to `Fr::from_be_bytes_mod_order`
- `hash_to_field_le` switched to `Fr::from_le_bytes_mod_order` (drops `BigUint` heap detour)
- `graph_from_raw` depth mismatch returns `GraphReadError::MissingSignal`

## Test migration

- Rewrote integration tests under `rln/tests/` with naming convention `test_<action>_<scope>_<variant>`:
- Deleted `tests/protocol.rs`, `tests/utils.rs`, `tests/poseidon_tree.rs`

## Benchmark
- Removed redundant benchmark files
- Standardized benchmark file, benchmark name convention
- Added benchmark operations for merkle tree for all ZerokitMerkleTree trait

## TODO
- Added a bunch of TODO(PR11) to handle in the next PR


rln/benches/poseidon_tree_benchmark.rs
sydhds, line 1: Just wondering but were these merkle tree bench moved somewhere else?

rln/ffi_c_examples/type_serialization.c
sydhds, line 1: Just wondering, do we provide mixed serialization in C API?

rln/src/circuit/mod.rs
```diff
@@ -202,52 +206,91 @@ pub fn graph_from_raw(
...
-pub fn default_graph_multi() -> &'static Graph {
+pub fn default_graph_multi() -> &'static Arc<Graph> {
     &GRAPH_MULTI
 }
 
-// Loads default Single zkey
-#[cfg(not(target_arch = "wasm32"))]
-pub fn default_zkey_single_v3() -> &'static Arc<Zkey> {
-    &ARKZKEY_SINGLE
+/// Secret field-element wrapper zeroized on drop.
```
sydhds: After thinking, I think this might be moved to another file. Better keep the mod.rs file as small as possible

rln/src/protocol/proof.rs
```diff
@@ -1,980 +1,123 @@
...
-    // Convert it to field elements
-    let mut field_elements = vec![];
-    for w in calculated_witness.into_iter() {
-        let w = if w.sign() == num_bigint::Sign::Minus {
-            // Need to negate the witness element if negative
-            modulus.into()
-                - w.abs()
-                    .to_biguint()
-                    .ok_or(ProtocolError::BigUintConversion(w))?
+/// Computes the Merkle tree root from identity credentials and Merkle membership proof.
```
sydhds, line 16: Please add a TODO because I think this should be added to a MerkleProof type or something similar

rln/tests/public.rs
```diff
@@ -1,33 +1,36 @@
...
+    use ark_std::UniformRand;
+    use rand::{rngs::ThreadRng, thread_rng, Rng};
     use rln::prelude::*;
     use serde_json::{json, Value};
+    use zerokit_utils::merkle_tree::{ZerokitMerkleProof, ZerokitMerkleTree};
 
-    fn fq_from_str(s: &str) -> Fq {
-        Fq::from_str(s).unwrap()
-    }
+    const NO_OF_LEAVES: usize = 256;
```
sydhds: LEAVES_COUNT?

rln/tests/public.rs
```diff
@@ -42,1785 +45,618 @@ mod test {
...
-            let identity_index_new = rln.leaves_set();
-            rln.set_next_leaf(rate_commitment_new).unwrap();
+    #[test]
+    fn test_groth16_proof_hardcoded_multi() {
+        let rln = RLNBuilder::stateless()
+            .graph(default_graph_multi().clone())
+            .zkey(default_zkey_multi().clone())
+            .build();
+
+        let snarkjs_proof = json!({
```
sydhds, line 221: Should we add a mention of how this was generated?

sydhds (changes requested):

rln/ffi_c_examples/type_serialization.c
vinhtc27: Yes, it's all there. You can see it in the rln.h file after running the generate command.
With Nim, I manually defined all functions in rln.nim because it doesn't use separate header files like C.

rln/benches/poseidon_tree_benchmark.rs
vinhtc27: I removed the benchmarks for FullMerkleTree and OptimalMerkleTree since they're kind of redundant cuz benchmarks already exist in the utils module.

I'm planning to move the entire pmtree implementation, including the Sled wrapper, into the rln module so we can properly showcase pmtree as a custom tree implementation. That way, users can just follow the same pattern for their own custom trees. 

And I've just added a benchmark file for pmtree, mirroring how the other tree benchmarks are set up in the utils module.

rln/tests/public.rs
vinhtc27: Should be like "LEAF_COUNT"

rln/benches/poseidon_tree_benchmark.rs
sydhds: Maybe we need a reflexion here as the tree implementation could be in a separate crate. If someone is using stateless mode, he might want to skip the tree implementations and use its own. Wdyt?

rln/src/circuit/mod.rs
sydhds: Reopen as IdSecret is still in mod.rs no?

sydhds (changes requested):

rln/benches/poseidon_tree_benchmark.rs
vinhtc27: Right now, the Sled wrapper of the PmTree lives in utils module, while the ZerokitMerkleTree trait is implemented by PmTree separately in the rln module.

I think there are two possible directions:

1. Move the Sled wrapper out of utils and into the rln module, keeping it alongside the ZerokitMerkleTree implementation. This is similar to what I mentioned before: it could serve as a reference/example implementation that we document, while allowing other users to plug in their own persistence layer if needed.

2. Move the ZerokitMerkleTree implementation into utils instead. In that setup, utils would contain the complete tree-related implementation (all 3 trees), along with benchmarks and tests, while the rln module would simply consume it. This would create a cleaner separation of concerns and align with what you suggested above.

Personally, I think either approach works. It mostly comes down to whether we want the tree implementation details to live closer to RLN, or keep everything tree-related consolidated in utils.

I've already added all the TODOs (PR11) to handle the remaining Merkle tree work in the next PR, so we can agree on the direction here and I'll implement it there.

WDYT? Which one do you prefer?

sydhds: Not sure now but all good to have this resolved in PR 11

sydhds (approved):

Issue #11, opened 2022-03-15
Add CI


oskarth:
Basic `cargo test` and possibly clippy.

Requires fixing up a few tests too.

oskarth:
Done in https://github.com/vacp2p/zerokit/commit/83ef5ae3c34259cad79fe7ccaa050ae58dc22f0c

Including all packages and with linting.

Not going to block merging on CI yet and will continue to push to master for now, but now it is at least setup for future PRs etc.

PR #421, opened 2026-06-25
refactor(PR #11): rework Merkle tree layer and error type, remove sled and pmtree utilities, migrate to pmtree v3, fix override_range bug, generic PmTree<D,H> struct for any persistent backend

### LINKED ISSUE #343: MerkleTree + too many users
Trying to add a new user in pmTree and to check if there are too many users: 

```rust

match self.merkle_tree.set(new_index, rate_commit) {
  Err(e) => {
    if matches!(e, ZerokitMerkleTreeError::PmtreeErrorKind(PmtreeErrorKind::TreeError(TreeErrorKind::IndexOutOfBounds))) {
      println!("Cannot add more users, merkle tree is full...");
    } else {
      println!("Other error
    }
  },
  _ => todo!(),
}
```

This is very complicated and rely on PmtreeErrorKind (so this code will fail to compile if we switch to another merkle tree impl).

vinhtc27:
## Changes: All the changes related to merkle tree impl in Zerokit
- Moved the sled-backed PmTree into rln module, utils module is now apure, storage-agnostic Merkle-tree crate with no sled/pmtree dependency.
- Make PmTree generic over the pmtree Database backend via a PmTreeBackendConfig trait (Sled impl by default). Allow any custom impl for persistent layer (RocksDB, PostgreSQL,...) with minimal impl via Database trait.
- ZerokitMerkleTree now has an associated `Error` type. The in-memory backends keep ZerokitMerkleTreeError, de-overloaded to one variant per condition; PmTree returns a structured PmTreeError that separates tree-logic errors (MerkleTree) from storage-backend faults (Backend), and pre-validates domain conditions so the same misuse yields the same variant on every backend. Callers match flat variants instead of nested enums ([#343](https://github.com/vacp2p/zerokit/issues/343)).
- Adopted pmtree 3.0.0: PmTree writes (set/delete/update_next/override_range) are now crash-atomic ([PR](https://github.com/vacp2p/pmtree/pull/6)).
~- PmTree is now generic over the hash (PmTree\<H\> instead of a hardcoded PoseidonHash), so a future PoseidonHash2 can plug in.~ (Generic over both persistent backend anh hash (PmTree\<D, H\>)
- Fixed override_range: replaced the per-tree span logic and its separate pre-validation pass (validate_override_range_inputs / EmptyIndicesPolicy in override_range_validation.rs, now deleted) with one shared default trait method — write the new leaves from start, reset only the delete-indices the write doesn't already cover (writes win on overlap), validate everything before touching the tree. 
- Fixed a bug where leaves landed at the wrong indices. Full/Optimal use the default; PmTree overrides it to apply the whole change in a single atomic commit.
- Tests: value + cache + validation coverage for override_range across all three merkle tree.
- Add TODO marked the RLN-math sites still calling Poseidon directly (id-commitment, nullifier, root recompute) for a later generic hash trait.
- Fixed silence delete opt in Full/OptimalMerkleTree.


rln/src/protocol/serialize.rs
```diff
@@ -179,6 +179,20 @@ impl CanonicalDeserializeBE for Vec<Fr> {
     }
 }
 
+impl CanonicalSerializeBE for [u8] {
+    type Error = SerializationError;
+
+    fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
+        writer.write_all(&serialize_usize_be(self.len()))?;
```
sydhds: So it seems here, the len (usize) is seriliazed as u64. Why not using directly u64 here?

rln/src/protocol/proof.rs
```diff
@@ -132,14 +108,22 @@ pub struct RLNProofValuesSingle {
...
+// until a real second hash + circuit land.
 impl From<&RLNWitnessInputSingle> for RLNProofValuesSingle {
     fn from(w: &RLNWitnessInputSingle) -> Self {
-        let root = compute_tree_root(
-            &w.identity_secret,
-            &w.user_message_limit,
-            &w.path_elements,
-            &w.identity_path_index,
-        );
+        let mut to_hash = [*w.identity_secret.clone()];
```
sydhds: Can't we implement the Hasher trait for IdSecret and use it in here and also for From<&RLNWitnessInputMulti> (and maybe other places?

rln/src/pm_tree.rs
```diff
@@ -1,76 +1,84 @@
...
-}
+const MAX_DEPTH: usize = 31;
 
-pub struct PmTreeProof {
-    proof: pmtree::tree::MerkleProof<PoseidonHash>,
-}
+pub type PmTreeMode = Mode;
 
-pub type FrOf<H> = <H as Hasher>::Fr;
+pub type FrOf<H> = <H as ZerokitHasher>::Fr;
```
sydhds: Just wondering but everytime I see this, I need to remember and find it. Could we remove this?

rln/src/pm_tree.rs
```diff
@@ -1,76 +1,84 @@
...
-        // TODO(PR11): change to hash_pair for this trait to use poseidon_hash_pair for PoseidonHash (new PR in vacp2p_pmtree)
-        poseidon_hash(inputs)
+    fn hash_pair(left: Self::Fr, right: Self::Fr) -> Self::Fr {
+        <Self as ZerokitHasher>::hash_pair(left, right)
     }
 }
 
+pub struct PmTree<H: Hasher> {
+    /// The underlying Merkle tree from the pmtree crate
+    tree: MerkleTree<SledDB, H>,
```
sydhds: Here we force the user to use SledDB as a backend but pmtree is about choosing a backend no? What if user wants a RocksDB backend here?

rln/src/pm_tree.rs
```diff
@@ -241,74 +259,76 @@ impl ZerokitMerkleTree for PmTree {
...
     }
 
-    fn set(
-        &mut self,
-        index: usize,
-        leaf: FrOf<Self::Hasher>,
-    ) -> Result<(), ZerokitMerkleTreeError> {
+    /// Returns the root of the subtree at level n and index
+    fn get_subtree_root(&self, n: usize, index: usize) -> Result<FrOf<Self::Hasher>, Self::Error> {
+        if n > self.depth() {
```
sydhds: I'm wondering but those checks should be in pmtree impl and not the adapter no?

rln/src/prelude.rs
```diff
@@ -1,13 +1,15 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+pub use pmtree::PmtreeError as PmTreeBackendError;
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use crate::circuit::{
     default_graph_multi, default_graph_single, default_zkey_multi, default_zkey_single,
 };
 #[cfg(not(target_arch = "wasm32"))]
-pub use crate::pm_tree::{FrOf, PmTree, PmTreeConfig, PmTreeProof};
+pub use crate::pm_tree::{FrOf, PmTree, PmTreeConfig, PmTreeError, PmTreeMode, PmTreeProof};
```
sydhds: Can we group the import here? And there is PmTreeError and PmTreeBackendError (which is PmtreeError). I'm kinda lost here...

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -83,9 +83,9 @@ where
         depth: usize,
         default_leaf: FrOf<Self::Hasher>,
         _config: Self::Config,
-    ) -> Result<Self, ZerokitMerkleTreeError> {
+    ) -> Result<Self, Self::Error> {
         if depth >= usize::BITS as usize {
-            return Err(ZerokitMerkleTreeError::InvalidDepth);
+            return Err(ZerokitMerkleTreeError::DepthTooLarge);
```
sydhds, line 88: DepthTooLarge is feeling a bit weird here. I'd rather keep InvalidDepth

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -182,112 +163,86 @@ where
...
         let index = self.capacity() + start - 1;
-        leaves.into_iter().for_each(|hash| {
-            self.nodes[index + count] = hash;
-            self.cached_leaves_indices[start + count] = 1;
-            count += 1;
-        });
-        if count != 0 {
-            self.update_hashes(index, index + (count - 1))?;
-            self.next_index = max(self.next_index, start + count);
+        for (offset, hash) in leaves.enumerate() {
```
sydhds: I get the use of enumerate here but why don't you inject the enumerate in before the for_each instead of rewriting with a for loop here? That would keep the number of changes lower...

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -182,112 +163,86 @@ where
...
+    }
 
-        self.set_range(start, set_values.into_iter())
+    /// Returns the indices of the leaves that are empty
+    fn get_empty_leaves_indices(&self) -> Vec<usize> {
+        self.cached_leaves_indices
+            .iter()
+            .take(self.next_index)
+            .enumerate()
+            .filter(|&(_, &v)| v == 0u8)
```
sydhds: filter_map here?

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -182,112 +163,86 @@ where
...
         Ok(())
     }
 
     /// Deletes a leaf at a certain index by setting it to its default value (next_index is not updated)
-    fn delete(&mut self, index: usize) -> Result<(), ZerokitMerkleTreeError> {
-        // We reset the leaf only if we previously set a leaf at that index
-        if index < self.next_index {
-            self.set(index, H::default_leaf())?;
-            self.cached_leaves_indices[index] = 0;
+    fn delete(&mut self, index: usize) -> Result<(), Self::Error> {
```
sydhds, line 232: You're changing the behavior of the code here. Previously there was no error on deleting an unset leaf. Why did you change this?

utils/src/merkle_tree/optimal_merkle_tree.rs
sydhds, line 1: Can we rename it to SparseMerkleTree here? Optimal merkle tree is kinda obscure naming no? FullMerkleTree should be renamed too wdyt?

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -138,28 +134,13 @@ where
...
-        }
-        Ok(self.nodes[self.capacity() + leaf - 1])
-    }
-
     /// Returns the root of the subtree at level n and index
-    fn get_subtree_root(&self, n: usize, index: usize) -> Result<H::Fr, ZerokitMerkleTreeError> {
+    fn get_subtree_root(&self, n: usize, index: usize) -> Result<H::Fr, Self::Error> {
         if n > self.depth() {
-            return Err(ZerokitMerkleTreeError::InvalidIndex);
+            return Err(ZerokitMerkleTreeError::LevelOutOfBounds);
```
sydhds, line 140: Level or Depth?

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -138,28 +134,13 @@ where
...
         }
         if index >= self.capacity() {
-            return Err(ZerokitMerkleTreeError::InvalidLeaf);
+            return Err(ZerokitMerkleTreeError::LeafIndexOutOfBounds);
```
sydhds, line 143: index here is not always a leaf no?

utils/src/merkle_tree/full_merkle_tree.rs
```diff
@@ -182,112 +163,86 @@ where
...
+        if leaf >= self.capacity() {
+            return Err(ZerokitMerkleTreeError::LeafIndexOutOfBounds);
+        }
+        self.set_range(leaf, once(hash))?;
+        self.next_index = max(self.next_index, leaf + 1);
+        Ok(())
     }
 
     /// Sets multiple leaves from the specified tree index
     fn set_range<I: ExactSizeIterator<Item = FrOf<Self::Hasher>>>(
```
sydhds, line 177: Do we need the collect here?

utils/src/merkle_tree/merkle_tree.rs
```diff
@@ -38,53 +38,87 @@ pub trait ZerokitMerkleTree {
...
         leaf: &FrOf<Self::Hasher>,
         merkle_proof: &Self::Proof,
-    ) -> Result<bool, ZerokitMerkleTreeError>;
-    fn set_metadata(&mut self, metadata: &[u8]) -> Result<(), ZerokitMerkleTreeError>;
-    fn metadata(&self) -> Result<Vec<u8>, ZerokitMerkleTreeError>;
-    fn close_db_connection(&mut self) -> Result<(), ZerokitMerkleTreeError>;
+    ) -> Result<bool, Self::Error>;
+    fn set_metadata(&mut self, metadata: &[u8]) -> Result<(), Self::Error>;
+    fn metadata(&self) -> Result<Vec<u8>, Self::Error>;
+    fn close_db_connection(&mut self) -> Result<(), Self::Error>;
```
sydhds: We should only use close() here as it feels a bit weird to call close_db_connection with an in memory merkle tree no? Also it should be marked as optional to call it; I guess the backend (for pmtree) is closed automatically on drop?

sydhds (changes requested):

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: It should be level because the getter method returns the subtree's root

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: WDYM? Its already use the ExactSizeIterator?

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: It's always a leaf

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: Previously it just do nothing in the overflow case on [zerokit](https://github.com/vacp2p/zerokit/blob/e077b5bb99ef3e5f4b8efbe3524c73070ed1bf45/utils/src/merkle_tree/optimal_merkle_tree.rs#L253), but the [pmtree](https://github.com/vacp2p/pmtree/blob/826c15e4c02f33a941130392a32ec98511577a6b/src/tree.rs#L199) crate doesn't allow overflow.
So I think it's better to align the logic across all Merkle tree implementations.

rln/src/prelude.rs
vinhtc27: PmtreeError is from pmtree crate, and i creaeted PmTreeError as wrapper.

```rust
#[derive(Debug, thiserror::Error)]
pub enum PmTreeError {
    #[error(transparent)]
    MerkleTree(#[from] ZerokitMerkleTreeError),
    #[error(transparent)]
    Backend(#[from] PmtreeError),
}
```

WDYT? Any idea to make this more clear?

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: Semantic is much more clear IMO, its better to have 2 seperate variant than single InvalidDepth
```rust
    /// The requested depth exceeds the supported maximum depth.
    #[error("Tree depth exceeds the supported maximum")]
    DepthTooLarge,
    /// The requested depth does not match the depth of the already existing (persisted) tree.
    #[error("Tree depth does not match the existing tree")]
    DepthMismatch,
```

rln/src/protocol/serialize.rs
sydhds: Re open as there is no comment nor any changes no?

rln/src/protocol/proof.rs
sydhds: Unresolved - no changes here...

rln/src/protocol/serialize.rs
vinhtc27: Im still resolving all your cmts, ill ping when this ready

utils/src/merkle_tree/optimal_merkle_tree.rs
vinhtc27: I don't think we should change it here. It required to rename alot of places for this.
I mean it's already well documented in the README as "Memory-efficient sparse storage using HashMap"

rln/src/protocol/proof.rs
vinhtc27: I added poseidon_hash_secret function to avoid these duplication, no need for impl the trait.

rln/src/pm_tree.rs
vinhtc27: I removed this one and use type FrOf from utils instead, its stadardized for all merkle tree.

rln/src/pm_tree.rs
vinhtc27: Yeah, pmtree also has that, but that's because it does its own checks for the external crate. This check is for ZerokitMerkleTreeError, which matches how the other in-memory trees handle errors.

rln/src/pm_tree.rs
vinhtc27: You right, this should be generic for any kind of persistent backend.

vinhtc27: I added the generics struct `PmTree<D: Database, H: Hasher>`, now user can have any persistent backend with minimal impl via Database trait.

rln/src/hashers.rs
```diff
@@ -52,6 +53,14 @@ pub fn poseidon_hash_pair(fr1: Fr, fr2: Fr) -> Fr {
         .expect("Two element input must be valid with supported round parameters")
 }
 
+/// Computes identity commitment from identity secret using Poseidon hash.
+pub fn poseidon_hash_secret(secret: &Fr) -> Fr {
```
sydhds, line 57: The function argument should be IdSecret here no?

rln/src/pm_tree.rs
```diff
@@ -1,85 +1,102 @@
...
-    /// Set to 0 if the leaf is empty and set to 1 in otherwise.
-    cached_leaves_indices: Vec<u8>,
-    // metadata that an application may use to store additional information
-    metadata: Vec<u8>,
-}
-
-pub struct PmTreeProof {
-    proof: pmtree::tree::MerkleProof<PoseidonHash>,
-}
+const MAX_DEPTH: usize = 31;
```
sydhds, line 23: This constant is already defined in pmtree crate no?

rln/src/pm_tree.rs
```diff
@@ -1,85 +1,102 @@
...
+    ZerokitMerkleTreeError,
 };
 
-use crate::{
-    circuit::Fr,
-    hashers::{poseidon_hash, PoseidonHash},
-};
+use crate::hashers::PoseidonHash;
 
 const METADATA_KEY: [u8; 8] = *b"metadata";
```
sydhds, line 19: This constant is already defined in pmtree crate no?

rln/src/pm_tree.rs
```diff
@@ -176,54 +193,71 @@ impl FromStr for PmTreeConfig {
...
+    fn tree_depth(&self) -> Option<usize> {
+        self.tree_depth
     }
+}
 
+impl<D, H> ZerokitMerkleTree for PmTree<D, H>
+where
+    D: Database,
+    D::Config: PmTreeBackendConfig,
+    H: ZerokitHasher + Hasher<Fr = FrOf<H>>,
```
sydhds, line 218: In terms of code reading, this line definitively looks confusing:

Here we have H defined as:

ZerokitHasher: Hasher trait from utils/src/merkle_tree/merkle_tree.rs
Hasher: Hasher trait from pmtree

Question here is why we need 2 traits? Because the Hasher trait from pmtree requires some serialization function. But both traits define default_leaf() & hash_pair() on the very same type.

I think we should mark this as a future PR to solve and have a proper design.

rln/src/protocol/serialize.rs
```diff
@@ -179,11 +168,25 @@ impl CanonicalDeserializeBE for Vec<Fr> {
...
+        VEC_LEN_BYTE_SIZE + self.len()
+    }
+}
+
 impl CanonicalSerializeBE for Vec<u8> {
     type Error = SerializationError;
 
     fn serialize<W: Write>(&self, mut writer: W) -> Result<(), Self::Error> {
-        writer.write_all(&serialize_usize_be(self.len()))?;
+        writer.write_all(&(self.len() as u64).to_be_bytes())?;
```
sydhds: Better use u64::try_from here + ? no?

utils/src/merkle_tree/full_merkle_tree.rs
sydhds: My bad here

utils/src/merkle_tree/full_merkle_tree.rs
sydhds: You're talking about overflow while I've mentionned unset leaf. Is that the same here? To me in terms of UX, it feels very weird to return an error in this case...

sydhds (changes requested):

utils/src/merkle_tree/full_merkle_tree.rs
vinhtc27: I mean, if the user calls the function for nothing, they're probably doing something wrong on their end. So an early error return could help in this case instead of making it a no-op.

rln/src/pm_tree.rs
vinhtc27: No it's zerokit thing, not in pmtree

rln/src/pm_tree.rs
vinhtc27: Yeah but it private so i redefined it here, i'll add /// mentioned it then

rln/src/hashers.rs
vinhtc27: Im using &Fr because keygen/seeded_keygen + the proof path pass IdSecret (deref-coerces), but extended_keygen/extended_seeded_keygen hold the secret as a plain Fr (returned raw alongside trapdoor+nullifier).
I added a TODO(PR12) to consider rename the IdSecret to SecretFr because trapdoor and nullifier in extended_keygen/extended_seeded_keygen should be cleanup too.

rln/src/prelude.rs
sydhds: Unresolve just to know what you have decided to do here?

vinhtc27: I removed the PmTreeBackendError alias and used PmtreeErrror directly

utils/src/merkle_tree/full_merkle_tree.rs
sydhds: Still feel weird but user can match error so I guess it's ok

rln/src/hashers.rs
sydhds: Good catch and I like the idea of renaming to SecretFr. Ok for TODO PR12

rln/src/pm_tree.rs
sydhds: So let's not make it private in pmtree. If the constant is proven useful, export it no?

rln-wasm/src/wasm_utils.rs
```diff
@@ -252,7 +252,7 @@ impl Identity {
         let seed_vec = seed.to_vec();
         let (identity_secret, id_commitment) = seeded_keygen(&seed_vec);
         Identity {
-            identity_secret,
+            identity_secret: *identity_secret,
```
sydhds, line 257: Add TODO to change that later as your leaking the secret here

rln/src/ffi/ffi_utils.rs
```diff
@@ -371,7 +371,7 @@ pub fn ffi_key_gen() -> repr_c::Vec<CFr> {
 #[ffi_export]
 pub fn ffi_seeded_key_gen(seed: &repr_c::Vec<u8>) -> repr_c::Vec<CFr> {
     let (identity_secret, id_commitment) = seeded_keygen(seed);
-    vec![CFr(identity_secret), CFr(id_commitment)].into()
+    vec![CFr(*identity_secret), CFr(id_commitment)].into()
```
sydhds, line 376: Add a comment here as you are leaking the secret here

sydhds (changes requested):

rln/src/pm_tree.rs
vinhtc27: I'll add a TODO for future change in pmtree

sydhds (approved):

PR #423, opened 2026-07-05
refactor(PR 12): add SecretFr, WasmSecretFr, FFI_SecretFr to zerokit, standardize type naming in Wasm and FFI

vinhtc27:
## Changes:
- rln: renamed IdSecret to SecretFr and apply the new type to nullifier, trapdoor, and key generation.
- rln-ffi: added opaque FFI_SecretFr, migrate to SecretFr, standardize FFI type naming, update C/Nim examples.
- rln-wasm: added WasmSecretFr, migrate to SecretFr, and standardize WASM type naming, update JS example.


rln/tests/proof.rs
```diff
@@ -481,7 +481,7 @@ mod test {
         let leaf_index = 3;
 
         let identity_secret_seed = hash_to_field_le(b"test-merkle-proof");
```
sydhds, line 483: Why not using `let mut` then avoiding the clone here?

rln/src/hashers.rs
```diff
@@ -53,14 +44,36 @@ pub fn poseidon_hash_pair(fr1: Fr, fr2: Fr) -> Fr {
...
+pub fn poseidon_hash_list(frs: &[Fr]) -> Result<Fr, PoseidonError> {
+    let hash = POSEIDON.hash(frs)?;
+    Ok(hash)
+}
+
+/// Computes the Poseidon hash of identity secret into a identity commitment.
+///
+/// The internal copy of the secret is zeroized after hashing.
+pub(crate) fn poseidon_hash_id_secret(secret: &SecretFr) -> Fr {
+    let mut to_hash = [**secret];
```
sydhds, line 59: deref here would be more readable wdyt?

rln/tests/keygen.rs
```diff
@@ -73,11 +73,11 @@ mod test {
     fn test_extended_keygen_relations() {
         let (trapdoor, nullifier, identity_secret, id_commitment) = extended_keygen();
 
-        let expected_identity_secret = poseidon_hash_pair(trapdoor, nullifier);
-        let mut to_hash = [identity_secret];
+        let expected_identity_secret = poseidon_hash_pair(*trapdoor, *nullifier);
```
sydhds, line 76: Why not use the hash pair for Secret Fr here ?

rln-wasm/src/wasm_rln.rs
```diff
@@ -100,20 +100,19 @@ pub struct WasmRLNWitnessInput(RLNWitnessInput);
 impl WasmRLNWitnessInput {
     #[wasm_bindgen(js_name = newSingle)]
     pub fn new_single(
-        identity_secret: &WasmFr,
+        identity_secret: &WasmSecretFr,
```
sydhds, line 103: I'm seeing this only now but is there a reason we use reference here everywhere? Because, for example, it requires to clone the secret here so almost no reason no?

sydhds (changes requested):

rln/tests/proof.rs
vinhtc27: Handled in next [PR 13](https://github.com/vacp2p/zerokit/pull/424).

rln/src/hashers.rs
vinhtc27: It still need * and then .deref so i guess it kinda the same, and also have to import some trait so i prefer this.

rln/tests/keygen.rs
vinhtc27: These change in the next [PR13](https://github.com/vacp2p/zerokit/pull/424)

rln-wasm/src/wasm_rln.rs
vinhtc27: Yes it has a good reason to do so.

WASM take & for all non-primitive types because with wasm-bindgen, a by-value param moves ownership, the JS object gets its internal pointer nulled after the call. JS has no move semantics, so callers don't expect this:

```javascript
const keys = WasmIdentityKeys.generate();
const secret = keys.getSecret();
const w1 = WasmRLNWitnessInput.newSingle(secret, ...); // if by-value: secret is consumed here
const w2 = WasmRLNWitnessInput.newSingle(secret, ...); // throws "null pointer passed to rust"
```
 With & the object stays alive and we clone once inside instead. Otherwise we need to provide clone() fn.
 
 The same idea apply for FFI too.

PR #424, opened 2026-07-06
refactor(PR 13): clean generic hash for all zerokit's module (prepare for future PoseidonHash2)

vinhtc27:
## Changes:
- replaced legacy Hasher/RLNHasher with generic ZerokitHasher
- refactored RLN hashing into generic Hasher facade
- made Groth16 backend and ZK proof generic over hashers
- simplified identity key API and add canonical serialization
- updated Merkle tree, PM tree, FFI, WASM and CLI to new hashing API
- removed legacy hash helpers and obsolete APIs
- aligned core, wasm and ffi to allow future PoseidonHash2 extend


utils/src/hasher.rs
```diff
@@ -0,0 +1,18 @@
...
+pub trait ZerokitHasher {
+    /// Type of the hashed elements, also used as the Merkle tree node type.
+    type Fr: Clone + Copy + Eq + Default + Debug + Display + FromStr + Send + Sync;
+
+    /// Hashes an arbitrary-length slice of field elements to a single field element.
+    fn hash(input: &[Self::Fr]) -> Self::Fr;
+}
+
+/// Shorthand for the element type of a [`ZerokitHasher`].
+pub type FrOf<H> = <H as ZerokitHasher>::Fr;
```
sydhds: Can we get rid of this? This is old code and make reading the code really hard ...

utils/src/hasher.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use std::{
+    fmt::{Debug, Display},
+    str::FromStr,
+};
+
+/// Defines the interface for a hash function over a prime field, used by all zerokit modules.
+pub trait ZerokitHasher {
+    /// Type of the hashed elements, also used as the Merkle tree node type.
+    type Fr: Clone + Copy + Eq + Default + Debug + Display + FromStr + Send + Sync;
```
sydhds: Please find another name because the name "Fr" conflicts with the ark type Fr here.
I don't think we need so many trait bounds no?

rln/src/hashers.rs
```diff
@@ -23,58 +23,63 @@ const ROUND_PARAMS: [(usize, usize, usize, usize); 8] = [
...
+/// The RLN hashing facade. All hashing in the crate goes through this one type.
 ///
-/// No panic or error is expected since the supported round parameters include the case of two elements.
-pub fn poseidon_hash_pair(fr1: Fr, fr2: Fr) -> Fr {
-    POSEIDON
-        .hash(&[fr1, fr2])
-        .expect("Two element input must be valid with supported round parameters")
-}
+/// For example, `Hasher::<PoseidonHash>::hash_pair(left, right)`.
+pub struct Hasher<H: ZerokitHasher<Fr = Fr>>(PhantomData<H>);
```
sydhds: Is there a reason to add trait bounds on a struct here? This is usually not a good idea to do so...

rln/src/hashers.rs
```diff
@@ -23,58 +23,63 @@ const ROUND_PARAMS: [(usize, usize, usize, usize); 8] = [
...
-/// The default Hasher trait used by Merkle tree implementation in utils.
-impl zerokit_utils::merkle_tree::Hasher for PoseidonHash {
-    type Fr = Fr;
+    /// Hashes a list of field elements.
+    pub fn hash_list(input: &[Fr]) -> Fr {
+        H::hash(input)
+    }
+
+    /// Computes the identity commitment `H(identity_secret)` from the identity secret.
+    pub fn compute_id_commitment(secret: &SecretFr) -> Fr {
```
sydhds: To me this is a big no, you are implementing keygen related function on a "Hashing" trait... We are mixing Hash stuff with Keygen stuff here

rln/src/hashers.rs
sydhds, line 1: To me, the "Hash" design could be improved as we force the user to use a `static POSEIDON` struct with no possibility for a user to change it. Also the static is behind a LazyLock (which requires some atomic sync each time we access it).

sydhds (changes requested):

rln/src/hashers.rs
vinhtc27: Currently, the code only exposes PoseidonHash and Hasher as public types, so users are required to use them as Hasher::\<PoseidonHash\>::some_method(). 
This also allows us to extend the API in the future by introducing PoseidonHash2 with a new static POSEIDON2 implementation.

And for LazyLock, its atomic operation overhead is negligible in this context, so it is not a concern here.

utils/src/hasher.rs
sydhds: Is there a particular need to the trait bounds here? You restricted the set but it still seems to me that we don't need any no?

rln/src/hashers.rs
sydhds: The idea here is to have a design ready for when we will introduce PoseidonHash2 or whatever hash method a Zerokit user would like to use (could be PoseidonHash with different round parameters for example). A carefully well designed API could allow that and also avoid the atomic operation; I agree this should not be a bottleneck in Zerokit but this seems easily avoidable here so why not doing it? Also the claim ("negligible") is a bold assumption because we don't know every case where Zerokit is used...

rln/src/hashers.rs
```diff
@@ -99,3 +82,27 @@ pub fn hash_to_field_be(signal: &[u8]) -> Fr {
 
     Fr::from_be_bytes_mod_order(&hash)
 }
+
```
sydhds, line 95: I've seen hash_to_field_le & hash_to_field_be here. 

* Can we rename the functions with a better name
* Add some doc why we have this function using Keccak256?

rln/src/protocol/zk.rs
sydhds, line 1: In general, maybe we could use where clause to ease the readability 👍 

```
impl<H: ZerokitHasher<Scalar = Fr>> RLNPartialZkProof for ArkGroth16Backend<H>
```

```
impl<H> RLNPartialZkProof for ArkGroth16Backend<H> where H: ZerokitHash<Scalar = Fr>>
```

Wdyt?

rln/src/protocol/witness.rs
```diff
@@ -7,25 +7,25 @@ use crate::{
...
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum RLNWitnessInput {
     Single(RLNWitnessInputSingle),
     Multi(RLNWitnessInputMulti),
 }
 
 impl RLNWitnessInput {
-    pub fn identity_secret(&self) -> &IdSecret {
+    pub fn identity_secret(&self) -> SecretFr {
```
sydhds, line 25: In order to avoid the clone here, can we make SecretFr Copy ? It would better match the use of Fr type?

rln/src/protocol/secret.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Computes the identity commitment `H(identity_secret)` from the identity secret.
+pub(crate) fn compute_id_commitment<H: ZerokitHasher<Scalar = Fr>>(
+    identity_secret: &SecretFr,
+) -> Fr {
+    let to_hash = Zeroizing::new([**identity_secret]);
+    H::hash(&*to_hash)
+}
+
+/// Computes the Shamir share slope `a_1 = H(identity_secret, external_nullifier, message_id)`.
+pub(crate) fn compute_share_slope<H: ZerokitHasher<Scalar = Fr>>(
```
sydhds, line 20: Naming is not very descriptive here no? Maybe we can find a better name here?

rln/src/protocol/secret.rs
```diff
@@ -0,0 +1,63 @@
...
+use zerokit_utils::hasher::ZerokitHasher;
+
+use crate::circuit::{Fr, SecretFr};
+
+/// Computes the identity commitment `H(identity_secret)` from the identity secret.
+pub(crate) fn compute_id_commitment<H: ZerokitHasher<Scalar = Fr>>(
+    identity_secret: &SecretFr,
+) -> Fr {
+    let to_hash = Zeroizing::new([**identity_secret]);
+    H::hash(&*to_hash)
```
sydhds, line 16: Why don't we have an impl of ZerokitHasher<Scalar = SecretFr> here and use it?

rln/src/protocol/keygen.rs
```diff
@@ -1,85 +1,139 @@
...
+#[derive(Debug, CanonicalSerialize, CanonicalDeserialize)]
+pub struct IdentityKeys {
+    pub(crate) identity_secret: SecretFr,
+    pub(crate) id_commitment: Fr,
+}
+
+impl IdentityKeys {
+    /// Generates a random RLN identity using a cryptographically secure RNG and the
+    /// protocol hash `H`.
+    pub fn generate<H: ZerokitHasher<Scalar = Fr>>() -> Self {
```
sydhds: If we go the generic way, should we also use a generic rand trait here?

rln/src/protocol/keygen.rs
```diff
@@ -1,85 +1,139 @@
...
+            id_commitment,
+        }
+    }
+
+    /// Generates a deterministic RLN identity from a seed using the protocol hash `H`.
+    ///
+    /// Uses ChaCha20 RNG seeded with the Keccak-256 hash of the input.
+    /// The same input always produces the same identity.
+    pub fn generate_seeded<H: ZerokitHasher<Scalar = Fr>>(seed: &[u8]) -> Self {
+        let mut rng = ChaCha20Rng::from_seed(chacha_seed(seed));
```
sydhds: It seems to me that you changed the original code here no (added the chacha_seed call) why?

sydhds (changes requested):

utils/src/hasher.rs
vinhtc27: Yes, each one has its use, you can check by removing any of them, the compiler will tell you to put it back
- Copy: tree nodes are passed around by value everywhere (root(), sibling reads, H::hash(&[left, right]))
- PartialEq: verify() compares the recomputed root with the stored one
- Default: ZerokitMerkleTree::default(depth) uses Scalar::default() as the empty-leaf value
- Debug: the tree and proof types derive Debug
- Send + Sync: the rayon parallel hashing path sends nodes across threads

rln/src/protocol/keygen.rs
vinhtc27: It the same think as before, it just create a common helper for both _seeded method for both struct

rln/src/hashers.rs
vinhtc27: I think the naming is good and shouldn't be renamed, but I'll add the documentation.

rln/src/protocol/witness.rs
vinhtc27: We can't, SecretFr is zeroize-on-drop, and Copy would create implicit bit-copies on the stack that no Drop (so no zeroize) ever.
And if you add derive macro Copy to SecretFr, compiler will show this - "the trait Copy cannot be implemented for this type; the type has a destructor Copy not allowed on types with destructors"

rln/src/protocol/secret.rs
vinhtc27: Because it conflict with the trait bound of Scalar, which require Copy as i explained above, and SecretFr cannot have Copy as i mention above too

rln/src/hashers.rs
vinhtc27: The current design is already extension-ready, just via types instead of instances, a future PoseidonHash2 is a new marker type with its own private static and its own circuit resources

rln/src/protocol/keygen.rs
vinhtc27: I mean, rand and rand_chacha are already the standard choices in the Rust ecosystem for a randomizer crate, so I think we should keep this simple.

rln/src/hashers.rs
sydhds: Yes it does but with some drawbacks like round parameters not visible for dev and as we mention requires a Lazylock. Can we at least try a new design here and choose at the end which one to keep?

utils/src/hasher.rs
sydhds: So at the end, you need those trait bounds because you are defining MerkleTree<H: ZerokitHasher> but this would not be the case if you would defined MerkleTree<T, H> where T: Copy, ... , H: Hasher<T> wdyt?

rln/src/protocol/keygen.rs
sydhds: Keeping it simple but we already have one generic here so adding another one will not complexify things a lot no?

rln/src/protocol/keygen.rs
sydhds: As yes, but I got confused with the name chacha_seed wich could be renamed like seed_for_chacha...

utils/src/hasher.rs
vinhtc27: Even if you make the type `T` separate, you still need `T` to match the `Scalar` type in in `fn hash()`:

```rust
pub trait ZerokitHasher {
    type Scalar;
    fn hash(input: &[Self::Scalar]) -> Self::Scalar;
}
```

And this would make the call site look like this:

```rust
let tree = FullMerkleTree::<Fr, PoseidonHash>::default(20)?;
```

So I'd say the current implementation is much cleaner for the call site.

```rust
let tree = FullMerkleTree::<PoseidonHash>::default(20)?;
```

rln/src/hashers.rs
vinhtc27: We have these backlog TODO:

```
// TODO(backlog): Generate these parameters.
// TODO(backlog): Implement automatic generation of round parameters.
```

The original idea was to generate these parameters at compile time and expose them as constants.
 
Could we tackle this in the next version that adds support for `PoseidonHash2`? We could also make the parameter generation generic so it works for both implementations later on.

The current implementation is already generic enough for callers and the upcoming `PoseidonHash2`.

utils/src/hasher.rs
sydhds: Why is `let tree = FullMerkleTree::<PoseidonHash>::default(20)?;` cleaner than `let tree = FullMerkleTree::<Fr, PoseidonHash>::default(20)?;`

The later shows, I want a MerkleTree storing `Fr` types and with `PoseidonHash` so it clearly states the intent of the dev. Also adding trait bounds for the trait Hasher makes no sense because the trait is meant for dev to impl it. 

Also another problem of those trait bounds, is that you cannot impl Hasher for SecretFr and that should not be the case no?

rln/src/hashers.rs
sydhds: So you plan to keep it like this for 3.0.0 and modify it for Zerokit 3.1.0 ?

vinhtc27: Yes, from the user standpoint, the api gonna be the same in term of usage, but we can have constant pre-generated PARAMS instead of current static one in Zerokit  3.1.0

utils/src/hasher.rs
vinhtc27: Ok so to resolve this:
- I removed all the bounds from `ZerokitHasher` (`type Scalar;` is now just bare). The tree types add the bounds they actually need with `where H::Scalar: ...`, you right about this one, cleaner split.
- `Scalar = SecretFr` still isn't allowed as a tree node, but that's intentional. Tree nodes are public commitments, while `SecretFr` is deliberately `!Copy` so secrets can't accidentally make their way into the generic tree/hash code.
- For `<T, H>`, each hasher already determines exactly one scalar type, so adding a separate `T` would just duplicate information the compiler already knows. I'd keep the single-parameter API.

sydhds: I still don't understand why you don't want to add the generic T? Any particular reason for this? It does improve the intent as already mentioned but you rejected it without any reason...

vinhtc27: The reason I prefer the single-parameter API is that `Scalar` isn't an independent choice, it's determined by `H` through the associated type. Any `<T, H>` design would still need `H: ZerokitHasher<Scalar = T>`, so `T` is always duplicate information that `H` already carries.

One consequence is that a two-parameter API allows expressing combinations that are invalid like:

```rust
let tree = FullMerkleTree::<WrongFr, PoseidonHash>::default(20)?;
```

But `PoseidonHash` already defines its scalar type:

```rust
impl ZerokitHasher for PoseidonHash {
    type Scalar = Fr;
}
```

We can definitely make PoseidonHash generic over any field element type as well, but that's outside the scope of this refactor, since entire codebase is still use arkworks types (`Fr`, and `SecretFr` wrapper).

sydhds (approved):

PR #425, opened 2026-07-08
refactor(PR 14): cleanup ffi, finalize public apis, rewrite documentation and readme, bump deps and version

vinhtc27:
## Changes:

- Bump dependencies.
- Use `pmtree` 3.0.1 which includes [PR #7](https://github.com/vacp2p/pmtree/pull/7).
- Bump the version of all modules to v3.0.0.
- Add Rust doc for all exposed types and functions in prelude.
- Simplify `public.rs`: construction goes through `RLNBuilder` only.
- Clean up FFI: unify function naming, pointer and result types.
- Redact `SecretFr` `Debug` output so secrets never leak into logs.
- Use `RLNMerkleProof` as the canonical Merkle proof type in witness builders.
- Share dependencies via `[workspace.dependencies]` with a single root `Cargo.lock`.
- Update all `README.md` files, examples, `CONTRIBUTING.md`, and other documentation.
- Rework error types: narrowest error enum per method, no `unwrap`/`panic!`/`unreachable!` in library code.
- Enforce structural invariants at `build()` (#426).


rln-wasm/src/wasm_utils.rs
```diff
@@ -98,6 +98,11 @@ impl WasmSecretFr {
     pub fn debug(&self) -> String {
         format!("{:?}", self.0)
     }
+
+    #[wasm_bindgen(js_name = equals)]
+    pub fn equals(&self, other: &WasmSecretFr) -> bool {
```
sydhds, line 103: PartialEq & Eq trait impl here?

rln/src/circuit/secret_fr.rs
```diff
@@ -53,12 +60,22 @@ impl Deref for SecretFr {
...
 }
 
-#[derive(Debug, Clone, Zeroize, ZeroizeOnDrop)]
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
 pub(crate) enum FrOrSecret {
     SecretFr(SecretFr),
     Fr(Fr),
 }
 
+impl fmt::Debug for FrOrSecret {
```
sydhds: You cannot derive Debug here (as SecretFr & Fr are both Debug)?

rln/src/circuit/types.rs
```diff
@@ -9,8 +9,11 @@ use ark_relations::r1cs::ConstraintMatrices;
 
 use crate::partial_proof::PartialProof as ArkPartialProof;
 
+/// Default maximum number of message-id slots supported by the multi circuit.
 pub const DEFAULT_MAX_OUT: usize = 4;
+/// Default Merkle tree depth.
 pub const DEFAULT_TREE_DEPTH: usize = 20;
+/// Byte size of a Groth16 proof in arkworks compressed form.
 pub const COMPRESS_PROOF_SIZE: usize = 128;
```
sydhds: I'm wondering if this constant should not a member of the trait  RLNZkProof wdyt?

rln/src/protocol/proof.rs
```diff
@@ -13,62 +13,73 @@ use crate::{
     hashers::Hasher,
 };
 
+/// The public values of an RLN proof, in either Single or Multi message-id mode.
 #[derive(Debug, Clone, PartialEq)]
 pub enum RLNProofValues {
+    /// Proof values for Single message-id mode.
```
sydhds: No need to overdocument here? What's the point here to doc every variant?

rln/src/protocol/proof.rs
```diff
@@ -13,62 +13,73 @@ use crate::{
...
     Single(RLNProofValuesSingle),
+    /// Proof values for Multi message-id mode.
     Multi(RLNProofValuesMulti),
 }
```
sydhds, line 23: Same here for each getter? This looks like this brings little value here no?

rln/src/public.rs
```diff
@@ -14,40 +14,83 @@ use crate::{
...
 #[derive(Debug, Clone)]
 pub struct Stateful<T> {
+    /// The Merkle tree owned by the RLN instance.
     pub tree: T,
 }
 
 impl<T> Stateful<T> {
+    /// Creates a new [`Stateful`] marker wrapping the given `tree`.
+    ///
+    /// # Examples
```
sydhds: Are we expecting dev to instantiate it here? And what's the value of this example?? Same for others below...

rln/src/public.rs
```diff
@@ -14,40 +14,83 @@ use crate::{
...
-/// Type-state marker for an RLN instance without merkle tree.
+/// Type-state marker for an [`RLN`] instance without a Merkle tree.
 #[derive(Debug, Clone)]
 pub struct Stateless;
 
+/// The RLN object.
+///
+/// It implements the methods required to update the internal Merkle tree and to generate and
+/// verify RLN zkSNARK proofs. The `State` type parameter is either [`Stateful`] (owns a Merkle
+/// tree) or [`Stateless`] (no tree); `ZkProof` is the zkSNARK backend.
```
sydhds: no tree is confusing at the very least. Stateless only means that the RLN object does not handle the merkle tree by itself no?

rln/README.md
```diff
@@ -14,90 +14,97 @@ This module allows you to:
...
+On native targets the example below uses the built-in default circuit resources, so no files need
+to be loaded.
+
+To use custom circuits, supply your own resources (see [Custom Circuit Compilation](#advanced-custom-circuit-compilation)):
 
 - `rln_final.arkzkey`: The proving key in arkzkey format.
-- `graph.bin`: The graph file built for the input tree size
+- `graph.bin`: The graph file built for the input tree size.
 
 ```rust
```
sydhds, line 38: I think it would be more useful to use an in mem merkle tree here as it would make the exampe more short wdyt?

rln/README.md
```diff
@@ -121,24 +128,22 @@ and it's used to prevent a RLN ZK proof generated for one application to be re-u
...
+- **Merkle Tree Implementations**: Multiple tree variants optimized for different use cases,
+  selected by type (there are no tree feature flags):
   - **Full Merkle Tree**: Fastest access with complete pre-allocated tree in memory.
-    Best for frequent random access (enable with `fullmerkletree` feature).
-  - **Optimal Merkle Tree**: Memory-efficient sparse storage using HashMap.
-    Ideal for partially populated trees (enable with `optimalmerkletree` feature).
+    Best for frequent random access (use the `FullMerkleTree` type).
+  - **Optimal Merkle Tree**: Memory-efficient sparse storage using `HashMap`.
+    Ideal for partially populated trees (use the `OptimalMerkleTree` type).
   - **Persistent Merkle Tree**: Disk-based storage with [sled](https://github.com/spacejam/sled)
```
sydhds: Update this to explain that the backend is configurable?

rln/README.md
```diff
@@ -121,24 +128,22 @@ and it's used to prevent a RLN ZK proof generated for one application to be re-u
...
-    for persistence across application restarts and large datasets
-    (enable with `pmtree-ft` feature).
+    for persistence across application restarts and large datasets (use the `PmTree` type).
 
 ## Building and Testing
```
sydhds, line 142: Would make sense to specify the package required for building (and optional ones for some targets like wasm32) then add a note of which OS is supported in the Makefile ?

rln/README.md
```diff
@@ -158,9 +163,6 @@ cargo make build
 
 # Test with default features
 cargo make test
```
sydhds, line 171: Appart from the deps installation not sure to see the interest of using `cargo make XXX` ?

rln/README.md
```diff
@@ -315,9 +317,7 @@ RLN provides C-compatible bindings for integration with C, C++, Nim, and other l
 The FFI layer is organized into several modules:
```
sydhds, line 325: We should add a note here about the limitations of the C API compared to the Rust API no?

rln/README.md
sydhds, line 1: The section Detailed Protocol Flow is not very informative I'd say. For instance, this line: 

```
**Rate Commitment**: Add commitment to a Merkle tree
```

This should be better developed and crystal clear even for a user with very little knowledge of RLN wdyt? 

Also for FFI related doc, I think the proof is always Groth16 so we should use that no?

utils/src/merkle_tree/merkle_tree.rs
```diff
@@ -12,11 +12,13 @@ use std::str::FromStr;
 use super::error::ZerokitMerkleTreeError;
 use crate::hasher::ZerokitHasher;
 
-/// Enables parallel hashing when there are at least 8 nodes (4 pairs to hash), justifying the overhead.
+/// Enables parallel hashing when there are at least 8 nodes (4 pairs to hash), justifying the
+/// overhead.
```
sydhds: This constants should be mentionned in the Readme no?

utils/Cargo.toml
```diff
@@ -13,11 +13,11 @@ bench = false
 
 [dependencies]
```
sydhds, line 15: It seems that multiple crate could be put in workspace and so only one version would be referenced like: 
* ark crates
* num-bigint
* zeroize

sydhds (changes requested):

rln/src/protocol/proof.rs
vinhtc27: I mean, it still has some information about None returns. And other method, I just kept it there for consistency.

rln/README.md
vinhtc27: We use it throughout the entire codebase, including in CI. 
For example, in WASM it's a useful shorthand for a really long, complicated command.

utils/Cargo.toml
vinhtc27: No, I don't think it should. Each module publishes its own crate, so we should keep the shared crate separate.

rln-wasm/src/wasm_utils.rs
vinhtc27: Deriving `PartialEq`/`Eq` for wrapper types like `FFI_SecretFr` or `WasmSecretFr` doesn't help because C and JavaScript don't support operator overloading, so they can't use Rust's `==` implementation, 

That's why I added a separate comparison function.

rln/README.md
vinhtc27: The Makefile script already checks the OS before installing any packages.

rln/src/circuit/types.rs
sydhds: It seems the constant has just been moved to serialize.rs. So if we defined this constant here, is this value returned in CanonicalSerialize::serialized_size ?

rln/README.md
sydhds: Yeah but here is the limitations:
* Only IOS & Linux Debian Ubuntu, so what about Windows / Fedora / other OS?
* The idea here is to specify the package required so the OS not supported can still do it

rln/README.md
sydhds: The new text is: "This registers the member and binds it to a per-epoch message budget." and I think this should be "... and binds him to a per epoch message budget" wdyt?

utils/src/merkle_tree/merkle_tree.rs
sydhds: No updates in Readme?

utils/Cargo.toml
sydhds: Not sure to understand your reasoning here? rln & zerokit_utils crate can share dependencies here so why not doing it? Especially since rln-wasm & rln-cli depends on both...

rln/src/circuit/types.rs
vinhtc27: I moved the constant to serialize.rs because it seemed like a good fit there. It's actually only used by CanonicalSerializeMixed::serialized_size.

utils/src/merkle_tree/merkle_tree.rs
vinhtc27: I had already added it under the `## Parallel Processing` section in rln module's README

utils/Cargo.toml
vinhtc27: I meant the [rln](https://crates.io/crates/rln/2.0.2/code/Cargo.toml) crate and the [utils](https://crates.io/crates/zerokit_utils/1.2.0/code/Cargo.toml) crate, which are published as two separate crates on crates.io

I think it's better not to share dependencies across crates in the same repository, since people browsing crates.io can see each crate's exact dependencies without needing to check GitHub.

rln/src/circuit/types.rs
sydhds: It seems your comment does not answer my question no?

rln/README.md
sydhds: "binds them" ? But here we are talking about the user no?

utils/src/merkle_tree/merkle_tree.rs
sydhds: In this section, I see no mention of this 8 CPU (or 8 cores) no? Also it would be better to use 8 CPU / 8 CPU Core instead of 8 nodes; it seems to me more explicit here...

utils/Cargo.toml
sydhds: rln crate and utils crate share some dependencies (for instance and non exhaustively, ark-ff and rayon both defined in rln/Cargo.toml & utils/Cargo.toml) so why not adding them only once in workspace Cargo.toml here? And why do we care about some display in crates.io here??

rln/src/public.rs
```diff
@@ -14,118 +14,119 @@ use crate::{
...
+    /// is reset to the default value (writes win on overlap). The internal `next_index` is updated
+    /// as in [`Self::set_leaves_from`].
+    ///
+    /// ## Example:
+    ///
+    /// ```
+    /// // Set two leaves from index 5 and clear the leaf at index 2 in a single commit.
+    /// rln.atomic_operation(5, vec![leaf0, leaf1], vec![2])?;
+    /// ```
     pub fn atomic_operation(
```
sydhds, line 130: Should we rename this to atomic_set? atomic_operation is not really meaningful no?

rln/src/protocol/witness.rs
```diff
@@ -2,82 +2,141 @@ use std::collections::HashSet;
...
     },
     error::{
         GenerateProofError, PartialWitnessInputError, WitnessInputMultiError,
         WitnessInputSingleError,
     },
 };
 
+/// A Merkle proof consisting of the path elements and
+#[derive(Debug, Clone, PartialEq, CanonicalSerialize, CanonicalDeserialize)]
+pub struct RLNMerkleProof {
```
sydhds, line 26: I think this would be better defining a trait here bc we have a generic merkle tree and thus we can expect to also have a generic merkle proof here wdyt?

sydhds (changes requested):

rln/src/circuit/types.rs
vinhtc27: Not sure what you mean, but we can just remove this constant and use `CanonicalSerialize::serialized_size(&self.proof, Compress::Yes)`

utils/src/merkle_tree/merkle_tree.rs
vinhtc27: It has nothing to do with CPU cores here. The threshold of 8 means that if a Merkle tree layer has at least 8 nodes to hash for the next layer, we use parallel hashing. Otherwise, sequential hashing is more efficient.

rln/src/public.rs
vinhtc27: I mean, even if we rename it to `atomic_set`, it still hides the fact that it can also delete leaves. `atomic_set_and_delete_leaves` is too long, and `override_range`, like the method in the Merkle tree, is also too vague. Users will still need to check the documentation anyway. I think that's unavoidable. Since this name is already used in another project, I don't think we should change it for such a complex tree operation.

rln/src/protocol/witness.rs
vinhtc27: We actually already have that trait, ZerokitMerkleProof in utils module, which every tree's proof implements. RLNMerkleProof is a different thing: it's just the plain data the witness stores (path_elements + identity_path_index), and there's a From<&P> impl bridging them, so any tree proof (even future ones) goes in via .merkle_proof(&tree_proof).
I did tried making the witness generic/trait-based, but it gets ugly fast, current impl is much cleaner.

rln/src/circuit/types.rs
sydhds: Maybe add a comment on the const about this equivalence? Or maybe define the const with this expression (not sure if serialized_size is const expr or not)

utils/Cargo.toml
sydhds: Maybe we can add the remaining ark-ec / ark-poly & ark-serialize into Cargo.toml workspace? That would make bumping the ark crates version easier (otherwise we might have version diff if we forget...)

rln/src/protocol/witness.rs
sydhds: Would you mind reverting this change and move it to another PR? Ideally we could merge this PR and continue this discussion in the next PR because there are already a lot of changes to see the impact of this feature

rln/src/circuit/types.rs
vinhtc27: I actually removed the constant entirely in the latest commits.

It now just delegates directly to Arkworks' `CanonicalSerialize::serialized_size(&self.proof, Compress::Yes)`, so it is the same as using the proof size constant before, without a separate constant.

I also changed the visibility of most constants to `pub(crate)`. We only need to expose these constants: `DEFAULT_MAX_OUT` and `DEFAULT_TREE_DEPTH`, every other constants are just internal stuff.

rln/src/protocol/witness.rs
vinhtc27: I'd prefer to keep it in for two practical reasons:

1. The actual API change is pretty small. Most call sites just changed from passing two arrays to `.merkle_proof(&tree_proof)` (or `RLNMerkleProof::new(...)` for the treeless case). The diff looks large because of the mechanical changes across witness constructors, FFI, WASM, and the examples.

2. Ekaterina's PR #426  is already based on this branch and has been rebased on top of these commits and had new commits on top that PR branch too.

I'd suggest we keep this PR as-is, then continue the API/design discussion afterward. That would gets us to the same end goal.

utils/Cargo.toml
sydhds: Maybe a good time to upgrade all ark crates to 0.6 no?

rln/src/protocol/witness.rs
```diff
@@ -398,17 +468,21 @@ pub struct RLNPartialWitnessInput {
 
 #[bon]
 impl RLNPartialWitnessInput {
+    /// Starts building a partial witness; call `build` to validate and construct it.
```
sydhds: validate is a misleading here as we have validate_against_graph here?

rln/src/protocol/witness.rs
sydhds: So can we try the generic version of the builder using the trait ZerokitMerkleProof as I see little to no value using the type RLNMerkleProof...

rln/src/protocol/witness.rs
sydhds, line 1: In point 2 of my comment, it seems that there was the idea of being able to create a witness and validate (validate_against_graph) it in the same call. Can we add this option in our builders?

sydhds (changes requested):
It seems to me that this is still not easy to understand when to use verify_with_signal & verify_with_roots. In terms of UX, a developer would not know which one to use... In which case, should we use verify_with_signal instead of verify_with_roots?

utils/Cargo.toml
vinhtc27: I tried this in a previous PR, but I broke the parallel feature because it doesn't allow a global thread pool in the browser.

rln/src/protocol/witness.rs
vinhtc27: I don't think we should add optional graph validation here for a few reasons:

- The graph check returns `GenerateProofError`, so all three witness error enums would need an additional variant that callers without a `Graph` would still have to handle, breaking the current clean separation of error enums.
- Method `generate_proof` has to perform the same validation anyway (witnesses can also come from deserialization), so doing it in `build()` would just duplicate the work.
- Most witness construction paths (deserialization, WASM/FFI, tests) don't have a `Graph` available.

Since `generate_proof` is the immediate next step and already returns the appropriate typed error, I'd keep the split of responsibilities as it is: `build()` validates structural invariants, while `generate_proof` validates graph consistency.

rln/src/protocol/witness.rs
vinhtc27: I did try the generic approach, but `RLNMerkleProof` still has to exist either way.

`RLNMerkleProof` is the serializable representation stored in the witness (used by deserialization, WASM/JS, FFI, tests, etc.). Those callers only have the raw path data, they don't have a Merkle tree or a type implementing `ZerokitMerkleProof`. That trait also carries `Hasher` and `compute_root_from`, which plain path data intentionally doesn't.

Because of that, making the builder generic WOULDN'T eliminate `RLNMerkleProof`; it would only change the builder signature. Tree proofs already flow in via `From<&P>`, so callers can simply do `.merkle_proof(&tree_proof)` today with essentially the same ergonomics. The generic version mainly adds a type parameter and lifetime to the builders without really simplifying the design.

If you have a cleaner way to express this with `ZerokitMerkleProof` that also removes the need for `RLNMerkleProof` (or otherwise simplifies the API), I'm happy to explore it in a follow-up PR. 

As it stands, though, this is the cleanest design I could come up with.

vinhtc27:
> It seems to me that this is still not easy to understand when to use verify_with_signal & verify_with_roots. In terms of UX, a developer would not know which one to use... In which case, should we use verify_with_signal instead of verify_with_roots?

I'll make the documentation clearer by explaining the purpose and usage of each public API, since we need to update it for the v3 release anyway.
https://lip.logos.co/anoncomms/raw/zerokit-api.html#proof-verification

utils/Cargo.toml
sydhds: Maybe report it in arkworks repo then?

rln/src/protocol/witness.rs
sydhds: Hum my idea was to use the types (like FullMerkleProof) implementing the ZerokitMerkleProof trait. If you provide a generic .merkle_proof in the builder then you could either: 
* store only the index_path / index_elements in the witness 
* Have a witness with generic merkle proof type

I think option 1 make sense (and is simpler) as we don't really need to extract the original type of the Merkle proof anymore (correct me here if this is wrong). With this, there is no need for RLNMerkleProof anymore no?

rln/src/protocol/witness.rs
sydhds: Make sense. So maybe the "real" issue here is that `validate_against_graph` is a bit unclear in terms of def UX especially since the function is `pub` but is also called auto. in generate_proof. Is there any reason a dev would want to call this manually?

sydhds:
> > It seems to me that this is still not easy to understand when to use verify_with_signal & verify_with_roots. In terms of UX, a developer would not know which one to use... In which case, should we use verify_with_signal instead of verify_with_roots?
> 
> I'll make the documentation clearer by explaining the purpose and usage of each public API, since we need to update it for the v3 release anyway. https://lip.logos.co/anoncomms/raw/zerokit-api.html#proof-verification

It seems to me that the doc has not yet changed no? Do you plan to update it in this PR?

rln/src/protocol/witness.rs
vinhtc27: It's actually `pub(super)`, not `pub(crate)` or `pub`. It has just enough visibility for `zk.rs` to call it, so users can't call it directly.

vinhtc27:
> It seems to me that the doc has not yet changed no? Do you plan to update it in this PR?

I'm working on the [PR](https://github.com/logos-co/logos-lips/pull/378) to update it, the one you already reviewed.

rln/src/protocol/witness.rs
vinhtc27: You're right that the witness only stores the flat path elements and path indices. The original tree proof type is discarded during `.build()` and never needed afterward.

The reason `RLNMerkleProof` still exists is that not every caller has a type implementing `ZerokitMerkleProof` trait. WASM, FFI, deserialization, and stateless users only have the raw path data, and that data cannot implement the trait because `ZerokitMerkleProof` also requires a `Hasher` and `compute_root_from()`.

Also `RLNMerkleProof` is not there to just preserve the original proof type. It is the canonical serializable representation of a Merkle proof that can be stored, serialized, and passed across core Rust and FFI (`FFI_RLNMerkleProof`) as a standalone value. 

Even with a generic builder, we would still need a type like this for those entry points. I think your idea could work for the pure Rust API, but we would still need a type like `RLNMerkleProof` as the platform-independent representation for those cross-platform entry points.

rln/src/protocol/witness.rs
sydhds: How I got confused by pub(super). Any reason to use this and not pub(crate) as the latter will not make it visible to the crate user no?

sydhds:
> > It seems to me that the doc has not yet changed no? Do you plan to update it in this PR?
> 
> I'm working on the [PR](https://github.com/logos-co/logos-lips/pull/378) to update it, the one you already reviewed.

Still it needs better doc in this PR as it will reflect in doc.rs no?

rln/src/protocol/witness.rs
sydhds: I've re read the wasm & ffi parts of the related code and I see why we need a RLNMerkleProof here in those cases.

It still seems to me that the RLNMerkleProof struct is unnecessary in pure Rust code so here is what I suggest: 
* Rust code: generic ZerokitMerkleProof, implement the trait for RLNMerkleProof (the trait would make the RLNMerkleProof a real and usable merkle proof type). Additionaly we could have From impl to convert to/from the other merkle proof (not sure if this is really useful or not)
* Wasm code: use the RLNMerkleProof as a real wasm type + use it in new_single / new_multi / new (partial). The RLNMerkleProof should have ser / dev functions to have feature parity with the FFI API
* FFI code: no changes as the witness new funcs already use FFI_RLNMerkleProof

With this changes we should have the best of both world: 
* Pure rust code use generic merkle proof
* Wasm & FFI have a real & concrete merkle proof type

Wdyt?

rln/src/protocol/witness.rs
vinhtc27: It's like the minimum visibility needed to expose it for zk.rs file to call. In general `pub` > `pub(crate)` > `pub(super)` > `private in term of visibility. 

But I'll change it to pub(crate) for consistency and to avoid confusion.

vinhtc27:
> Still it needs better doc in this PR as it will reflect in doc.rs no?

Updated the /// docs for the verify methods in public.rs.

rln/src/protocol/witness.rs
vinhtc27: I've implemented the WASM side of your suggestion. `WasmRLNMerkleProof` now wraps `RLNMerkleProof`, is used by `newSingle` / `newMulti` / partial witness, and has its own getters plus LE/BE ser/de, so it's now on par with `FFI_RLNMerkleProof`. The JS examples also construct and pass it around the same way the C examples do.

That's also why I'd prefer to keep `RLNMerkleProof` in core as a plain data type. Both the FFI and WASM types are just thin wrappers around it, so the wire format and serialization only need to live in one place instead of being duplicated. 

With one canonical type, unified serialization, and the existing `From` conversions already making merkle proofs ergonomic, I think it keeps things cleaner and more consistent.

sydhds: Yes you can keep the RLNMerkleProof type but let's go with generic type for the Rust API. with it the Rust dev have full control & great flexibility of the API that we are trying to provide...

rln-cli/Cargo.toml
```diff
@@ -20,10 +21,10 @@ name = "partial"
 path = "src/examples/partial.rs"
 
 [dependencies]
-rln = { path = "../rln", version = "2.0.1", default-features = false }
-zerokit_utils = { path = "../utils", version = "1.2.0", default-features = false }
-clap = { version = "4.6.1", features = ["cargo", "derive", "env"] }
-rand = "0.8.6"
+rand = { workspace = true }
+rln = { path = "../rln", version = "3.0.0", default-features = false }
```
sydhds, line 25: Just wondering if the version field here is required here? If the version is updated in rln crate (let's say 4.0) it will not compile anymore?

rln-wasm/tests/browser.rs
```diff
@@ -94,19 +94,20 @@ mod test {
...
+        let tree_merkle_proof: OptimalMerkleProof<PoseidonHash> =
+            tree.proof(identity_index).unwrap();
         let mut path_elements = VecWasmFr::new();
-        for path_element in merkle_proof.get_path_elements() {
+        for path_element in tree_merkle_proof.get_path_elements() {
             path_elements.push(&WasmFr::from(path_element));
         }
-        let path_index = Uint8Array::from(&merkle_proof.get_path_index()[..]);
+        let path_index = Uint8Array::from(&tree_merkle_proof.get_path_index()[..]);
+        let merkle_proof = WasmRLNMerkleProof::new(&path_elements, &path_index);
```
sydhds: You should use a From impl here? It seems to me that we've discussed about From impl to convert a OptimalMerkleProof / FullMerkleProof / PmMerkleProof to RLNMerkleProof no?

rln-wasm/tests/node.rs
```diff
@@ -50,19 +49,20 @@ mod test {
...
+        let tree_merkle_proof: OptimalMerkleProof<PoseidonHash> =
+            tree.proof(identity_index).unwrap();
         let mut path_elements = VecWasmFr::new();
-        for path_element in merkle_proof.get_path_elements() {
+        for path_element in tree_merkle_proof.get_path_elements() {
             path_elements.push(&WasmFr::from(path_element));
         }
-        let path_index = Uint8Array::from(&merkle_proof.get_path_index()[..]);
+        let path_index = Uint8Array::from(&tree_merkle_proof.get_path_index()[..]);
+        let merkle_proof = WasmRLNMerkleProof::new(&path_elements, &path_index);
```
sydhds: Same as previous (should use From impl here...)

sydhds (changes requested):

rln-cli/Cargo.toml
vinhtc27: Yes, the compiler will enforce that the version matches the current version of the workspace modules. 
It's not required, but it makes the dependency more explicit.

rln/src/protocol/witness.rs
vinhtc27: I actually prototyped this approach (using a generic `P: ZerokitMerkleProof` setter and implementing the trait for `RLNMerkleProof`).

It works, but I don't think it adds any value for the current API. With current `#[builder(into)]` and the `From<&P>` impl, the builder already accepts any `ZerokitMerkleProof` implementation, both current and future, using the exact same call:

```rust
.merkle_proof(&tree_proof)
```

So nothing changes for the caller. The only difference is the implementation behind the scenes.

Right now the RLNWitnessInput builder is pretty simple:

```rust
pub fn new_single(
    ...,
    #[builder(into)] merkle_proof: RLNMerkleProof,
    ...,
) -> Result<Self, WitnessInputSingleError> {
```

With the generic version, every RLNWitnessInput builder ends up with an extra type parameter and trait bounds:

```rust
pub fn new_single<P>(
    ...,
    merkle_proof: &P,
    ...,
) -> Result<Self, WitnessInputSingleError>
where
    P: ZerokitMerkleProof<Index = u8>,
    P::Hasher: ZerokitHasher<Scalar = Fr>,
{
```

On top of that, `RLNMerkleProof` has to implement the trait, which means we're attaching protocol-specific hashing to what is currently just a plain data type:

```rust
impl ZerokitMerkleProof for RLNMerkleProof {
    type Index = u8;
    type Hasher = PoseidonHash;

    fn compute_root_from(&self, leaf: &Fr) -> Fr { ... }
    ...
}
```

That works today, but once we support another hash, this impl either hardcodes the choice or forces us to make `RLNMerkleProof` generic over the hasher. That generic then starts flowing through the witness types, serde, and API boundaries. We end up with the same flexibility, but a lot more generics and trait bounds to carry around. Current design doesn't have that problem if we support another hash.

sydhds: Too bad we can't have the RLNMerkleProof as a "real" merkle proof type. I think our ZerokitMerkleProof trait is not that well designed but that will be for a futur PR.

So to ack this, I would rename RLNMerkleProof to a name reflecting this data type situation (because here a dev could be confused between all these merkle proof types and RLN prefix make it like it's the proof to use). Also I would make the visibility pub(crate) as this type is only intended to be used internally. Also doc should make it clear that this is a data type (and explain why we need it) + doc for From impl (mentioning the other Merkle tree proof types). Wdyt?

sydhds (changes requested):

rln/src/protocol/witness.rs
vinhtc27: - Added more detailed documentation for the struct and its `From` implementation.
- It should be `pub`, not `pub(crate)`. As mentioned in my previous comment, this is the inner type of `FFI_RLNMerkleProof` and `WasmRLNMerkleProof` so it use outside rln crate too.
- I'd prefer to keep the current name. The `RLN` prefix identifies the protocol's family of data types (`RLNWitnessInput`, `RLNProofValues`, `RLNProof`, and `RLNMerkleProof`), so renaming only this type would break that consistency. The boundary types (`FFI_RLNMerkleProof`, `WasmRLNMerkleProof`) and the C symbols (`ffi_rln_merkle_proof_*`) are also named after it. Since this name has already existed in previous versions of the FFI and most of zerokit user are using Nim FFI, I'd avoid create new name for it.

sydhds:
The doc comment on RLNMerkleProof is unclear, I would suggest something like this instead: 

```
A data type representing a merkle proof used in `RLNWitnessInput` builders function. It is used in a Stateless context (for instance in Wasm) and/or FFI. Rust users in Stateful context (e.g. with a tree handled in the RLN struct), will usually use the ZerokitMerkleProof structs available (OptimalMerkleProof / ...) as they have an auto From impl.
```

Wdyt?

vinhtc27:
> Wdyt?

Updated with a slightly more formal tone.

sydhds (approved):

PR #426, opened 2026-07-14
feat(rln): enforce structural invariants on deserialization

seemenkina:
Enforces the invariants that the constructors already check (`new_single`/`new_multi`, `from_witness`) when values are **deserialized**, not only when constructed. The derived `CanonicalDeserialize` validates each field in isolation and can't express cross-field/range rules, so these were previously skipped on the wire path.

## Why
A mismatched-length `RLNProofValuesMulti` (`ys`/`nullifiers`/`selector_used` of differing lengths) could be deserialized and then make `recover_secret` panic with an out-of-bounds index on `ys[i]` — a regression from v2.0.2, which rejected it. Witness invariants (non-zero limit, `message_id < limit`, matching path/selector lengths, etc.) were likewise unenforced on deserialize.

## What
- Added `validate()` to `RLNProofValuesMulti` and the three witness types, enforced on both the compressed and big-endian deserialize paths (and, for proof-values, defensively in `recover_secret`).
- New `SerializationError::InconsistentProofValueLengths`; surfaced from `recover_secret` via `RecoverSecretError::InvalidProofValues`.
- Tests added to `tests/serialize.rs`, `tests/proof.rs`, and a crate-internal module in `witness.rs`.


rln/tests/proof.rs
```diff
@@ -336,6 +336,28 @@ mod test {
...
+            selector_used: vec![true],
+        });
+        let other = RLNProofValues::Single(RLNProofValuesSingle {
+            y: Fr::from(3u64),
+            root: Fr::from(1u64),
+            nullifier: Fr::from(42u64),
+            x: Fr::from(5u64),
+            external_nullifier: Fr::from(9u64),
+        });
+        assert!(malformed.recover_secret(&other).is_err());
```
sydhds: Shoud we test for a specific error here?

rln/src/protocol/proof.rs
```diff
@@ -218,6 +218,15 @@ impl RLNProofValuesMulti {
             selector_used: w.selector_used.clone(),
         }
     }
+
+    /// Checks that `ys`, `nullifiers`, and `selector_used` all have the same length.
+    pub fn validate(&self) -> Result<(), SerializationError> {
```
sydhds: So nothing wrong about this function but in terms of UX, should we allow to build an invalid RLNProofValuesMulti by itself? So maybe we could remove the pub for fields and provide a new method that would use this validate func. Wdyt?

rln/src/protocol/serialize.rs
```diff
@@ -263,8 +263,18 @@ impl CanonicalDeserializeBE for Vec<bool> {
 impl Valid for RLNWitnessInput {
     fn check(&self) -> Result<(), ArkSerializationError> {
         match self {
-            RLNWitnessInput::Single(inner) => inner.check(),
-            RLNWitnessInput::Multi(inner) => inner.check(),
+            RLNWitnessInput::Single(inner) => {
```
sydhds: I see one problem here. It seems to me that we should manually impl CanonicalDeserialize and add this validate call in check otherwise a user can deser an RLNWitnessInputSingle manually and turn it into a RLNWitnessInput. Wdyt?

rln/src/protocol/proof.rs
seemenkina: I think that we can even make fields in RLNProofValuesMulti/RLNProofValuesSingle as pub crate 

because logically if we make fields as public is better to provide something to validate it, but if we hide validation, let's also hide the fields in the structure itself, wdyt?

rln/tests/proof.rs
seemenkina: we can, I'll do it

rln/src/protocol/serialize.rs
```diff
@@ -307,15 +317,20 @@ impl CanonicalDeserialize for RLNWitnessInput {
...
+                ENUM_TAG_SINGLE => RLNWitnessInput::Single(
+                    RLNWitnessInputSingle::deserialize_with_mode(&mut reader, compress, validate)?,
+                ),
+                ENUM_TAG_MULTI => RLNWitnessInput::Multi(
+                    RLNWitnessInputMulti::deserialize_with_mode(&mut reader, compress, validate)?,
+                ),
+                _ => return Err(ArkSerializationError::InvalidData),
+            };
+        if let Validate::Yes = validate {
+            value.check()?;
```
vinhtc27, line 440: This is redundant when `validate == Validate::Yes` since the derived `deserialize_with_mode` for `CanonicalDeserialize` already calls `inner.check()` internally.

Same for the other two `if let Validate::Yes = validate` blocks with `RLNPartialWitnessInput` and `RLNProofValues`.

rln/tests/serialize.rs
vinhtc27, line 1: Let's add testcase for both LE and BE here

rln/src/protocol/serialize.rs
vinhtc27, line 1: I think we're making this more complicated than it needs to be by pushing the invariant checks into the serde layer.

I do like the idea of separating out `validate()`, but I'm not convinced we should require it to be called during deserialization.

For BE this is pretty clean since it's our own trait and error type, but on the LE side `ark`'s `Valid`/`CanonicalDeserialize` force us into handwritten impls and a generic `InvalidData` that throws away all the useful error details. That's where most of the complexity in this diff is coming from.

Stepping back, I'm not sure we actually need `deserialize` itself to reject invalid values. In practice, we're deserializing bytes that came from values built through the validating builders anyway. And with the fields being `pub(crate)` (Sylvain's comment), there's no way for code outside the crate to manually construct an invalid value.

The only case left is someone intentionally feeding us crafted bytes. For that, we can just make `validate()` public calling it from the builders (already have this in builder pattern), and also call it right before `generation_proof`, `verify` and in `recover_secret`. That way invalid values are still caught before they're actually used, but we keep the full typed error instead of collapsing everything into `InvalidData`.

That would also let us keep the derived impls and avoid all the extra LE serde code.

vinhtc27 (changes requested):

rln/src/protocol/proof.rs
sydhds: Yeah totally

rln/src/protocol/serialize.rs
sydhds: I'm more in favor (for performance reason) of not letting user having non valid struct than calling validate everywhere in generation_proof / verify / recover_secret

sydhds (approved):

rln/src/protocol/serialize.rs
seemenkina: fixed

rln/tests/serialize.rs
seemenkina: added

vinhtc27 (approved):

sydhds (approved):

PR #433, opened 2026-08-06
feat: extended round parameters for PoseidonHash

### LINKED ISSUE #50: Generate Poseidon round parameters and constants

vinhtc27:
## Changes:
- Updated rln-wasm README file.
- Extended round parameters for PoseidonHash.
- Created single centralize BN254_ROUND_PARAMS in utils.
- Created testcase with cross check with circomlibjs real values.
- Declared skip-benchmark and run-coverage labels in labels.yml so the label syncer stops pruning them.
- Removed Throughput from poseidon benchmark: criterion-compare-action cannot parse it (boa-dev/criterion-compare-action#22), skip-benchmark label added to bypass the one-time crash on this PR.


rln/src/hashers.rs
```diff
@@ -98,22 +86,60 @@ pub fn hash_to_field_be(signal: &[u8]) -> Fr {
 mod test {
     use super::*;
 
+    #[test]
+    fn test_poseidon_hash_against_circomlib() {
+        // Real outputs from circomlibjs `poseidon([1n, ..., n])` for arity `1..=16`.
+        // Pins circom interop and `SKIP_MATRICES = 0` for every width in `BN254_ROUND_PARAMS`.
+        let real_circomlibjs_outputs = [
```
seemenkina, line 102: can you mention particular file where you take it?

vinhtc27: done

rln/src/hashers.rs
```diff
@@ -98,22 +86,69 @@ pub fn hash_to_field_be(signal: &[u8]) -> Fr {
 mod test {
     use super::*;
 
+    #[test]
+    fn test_poseidon_hash_against_circomlib() {
+        // Real outputs from circomlibjs `poseidon([1n, ..., n])` for arity `1..=16`, not copied
+        // from any file: each value was generated by running circomlibjs directly:
```
seemenkina:
```suggestion
        // Real outputs from circomlibjs `poseidon([1n, ..., n])` for arity `1..=16`;
        // each value was generated by running circomlibjs directly:
```

seemenkina: unnecessary comment as you next provide generation script

seemenkina (approved):

PR #436, opened 2026-08-13
nix: add rln-stateless output; disable defaults when features are set

richard-ramos:
> **Not for merge.** Draft opened for visibility only. `master` (v3.0.0) removed the `stateless` feature entirely, so this fix does not apply there as-is. We'll re-do the equivalent against v3 once we migrate our consumers off `ffi_rln_new`'s v2 zero-argument form. Filing here so the v2.x nix packaging is at least reproducible in the meantime.

The `stateless` feature exposes the zero-argument form of `ffi_rln_new`, which downstream Nim/C bindings (e.g. `mix-rln-spam-protection-plugin`) rely on. There is currently no packaged way to consume it — every caller has to fork this flake to add `override { features = "stateless"; }`.

### Changes

- **`flake.nix`**: expose `rln-stateless = buildRln.override { features = "stateless"; }` alongside `rln`, so callers can pull it in directly.
- **`nix/default.nix`**: when the caller passes `features`, add `--no-default-features`. Without this, `--features=stateless` still pulls in `parallel` (the current default), and the stateless build fails to compile with `parallel` enabled.
- **`cargoHash`** bumped: the previous value no longer matches the vendor set that `rustPlatform.buildRustPackage` computes on the current nixpkgs 25.11 pin.

### Verification

\`\`\`
nix build .#rln           # unchanged output, still works
nix build .#rln-stateless # produces librln.{a,so} without parallel
\`\`\`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,6 +51,10 @@
       in rec {
         rln = buildRln;
 
+        # mix-rln-spam-protection-plugin's Nim binding calls the zero-arg
+        # form of ffi_rln_new, which zerokit exposes only under `stateless`.
+        rln-stateless = buildRln.override (old: { features = "stateless"; });
+
         rln-linux-arm64 = buildRln {
           target-platform = "aarch64-multiplatform";
           rust-target = "aarch64-unknown-linux-gnu";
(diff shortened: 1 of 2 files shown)


richard-ramos:
Closing this draft because the current Mix and Delivery integration uses the shared RLN module and does not need this change merged upstream. This PR only packages Zerokit v2 for the legacy embedded provider, and it does not apply to current master, which removed the stateless feature. We are keeping the fork branch because existing legacy builds still reference it.

